PR #1, opened 2021-01-04
Bootstrap blynn-compiler up to marginally

stikonas:
Further bootstrap needs some more thought since kaem does not support IO redirection.


fosslinux:
Thanks!

PR #5, opened 2021-01-11
Build sed 1.18.

stikonas:


fosslinux:
Merged, even though github dosen't think so.

PR #7, opened 2021-01-12
add chroot-mode, so it works without qemu/kernel

bittorf:
Like proposed in IRC from user 'mihi'.
This makes it possible to testrun it without a kernel/qemu.

### DIFF
--- a/rootfs.sh
+++ b/rootfs.sh
@@ -115,7 +115,7 @@ find . | cpio -H newc -o | gzip > initramfs.igz
 
 # Run
 if [ "$QEMU_CMD" = 'chroot' ]; then
-	PATH="/after/bin:$PATH" chroot . /init
+	PATH="/after/bin:$PATH" sudo chroot . /init
 else
 	$QEMU_CMD -enable-kvm \
 	    -m "${QEMU_RAM:-8G}" \
--- a/rootfs.sh
+++ b/rootfs.sh
@@ -115,7 +115,7 @@ find . | cpio -H newc -o | gzip > initramfs.igz
 
 # Run
 if [ "$QEMU_CMD" = 'chroot' ]; then
-	PATH="/after/bin:$PATH" sudo chroot . /init
+	sudo PATH="/after/bin:$PATH" chroot . /init
 else
 	$QEMU_CMD -enable-kvm \
 	    -m "${QEMU_RAM:-8G}" \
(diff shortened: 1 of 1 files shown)


rootfs.sh
```diff
@@ -114,13 +114,18 @@ cd tmp
 find . | cpio -H newc -o | gzip > initramfs.igz
 
 # Run
-${QEMU_CMD:-qemu-system-x86_64} -enable-kvm \
-    -m "${RAM:-8G}" \
-    -nographic \
-    -no-reboot \
-    -kernel ../../kernel -initrd initramfs.igz -append console=ttyS0
+if [ "$QEMU_CMD" = 'chroot' ]; then
+	PATH="/after/bin:$PATH" sudo chroot . /init
```
stikonas, line 118: maybe swap `sudo` and `PATH`
```suggestion
	sudo PATH="/after/bin:$PATH" chroot . /init
```

rootfs.sh
```diff
@@ -1,14 +1,14 @@
...
-RAM=$2
+QEMU_CMD="${1:-qemu-system-x86_64}"	# or 'chroot'
+QEMU_RAM="${2:-8G}"
 
 pushd sysa
 
 # SYSTEM A
 
-# Setup tmp 
+echo "trying to mount ramdisk/tmpfs: size 8G"
```
fosslinux, line 11: Don't think this is necessary...

rootfs.sh
```diff
@@ -114,13 +114,18 @@ cd tmp
...
 # Run
-${QEMU_CMD:-qemu-system-x86_64} -enable-kvm \
-    -m "${RAM:-8G}" \
-    -nographic \
-    -no-reboot \
-    -kernel ../../kernel -initrd initramfs.igz -append console=ttyS0
+if [ "$QEMU_CMD" = 'chroot' ]; then
+	PATH="/after/bin:$PATH" sudo chroot . /init
+else
+	$QEMU_CMD -enable-kvm \
```
fosslinux, line 120: Please change `$QEMU_CMD` to `"${QEMU_CMD}"`.

fosslinux (changes requested):

rootfs.sh
stikonas: Is this necessary? `shellcheck` doesn't complain about this and I think there is no risk of escaping here.

PR #8, opened 2021-01-12
add chroot-mode, so it works without qemu/kernel

bittorf:
rebased intop of 98058e6b70ba4e28839c0b4898d7049c172ea0d4

### DIFF
--- a/rootfs.sh
+++ b/rootfs.sh
@@ -9,7 +9,6 @@ pushd sysa
 # SYSTEM A
 
 # Setup tmp
-echo "trying to mount ramdisk/tmpfs: size 8G"
 mkdir -p tmp/
 sudo mount -t tmpfs -o size=8G tmpfs tmp
 
@@ -134,5 +133,4 @@ fi
 cd ../..
 
 # Cleanup
-echo "trying to unmount ramdisk/tmpfs, abort with CTRL+C"
 sudo umount sysa/tmp
--- a/rootfs.sh
+++ b/rootfs.sh
@@ -99,8 +99,9 @@ get_file() {
     else
 	bname=$(basename "$url" ".tar.${ext}")
     fi
-    cp "${bname}.kaem" tmp/after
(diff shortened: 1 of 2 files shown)


rootfs.sh
```diff
@@ -98,8 +99,9 @@ get_file() {
     else
 	bname=$(basename "$url" ".tar.${ext}")
     fi
-    cp "${bname}.kaem" tmp/after
-    cp "../sources/$(basename "$url")" tmp/after
+
+    [ -f "${bname}.kaem" ] && cp -v "${bname}.kaem" tmp/after
+    cp -v "../sources/$(basename "$url")" tmp/after
```
stikonas, line 104: you can drop this now. I've reworked this part of `get_file` function.

rootfs.sh
```diff
@@ -119,13 +121,18 @@ cd tmp
 find . | cpio -H newc -o | gzip > initramfs.igz
 
 # Run
-${QEMU_CMD:-qemu-system-x86_64} -enable-kvm \
-    -m "${RAM:-8G}" \
-    -nographic \
-    -no-reboot \
-    -kernel ../../kernel -initrd initramfs.igz -append console=ttyS0
+if [ "${QEMU_CMD}" = 'chroot' ]; then
+	sudo PATH="/after/bin:${PATH}" chroot . /init
```
stikonas, line 125: Maybe add a comment that this is a bug in old version of kaem but is not required with new kaem?

rootfs.sh
bittorf: but your change does not work/breaks.

IMHO better is:
```
for FILE in "${base}."*; do test -f "$FILE" && cp "$FILE" tmp/after; done
```
What do you think?

rootfs.sh
bittorf: if this is not needed anymore, i can just drop die PATH export.
are you sure?

fosslinux:
Merged.

PR #9, opened 2021-01-12
Copy all types of build scripts into sysa, not just kaem.

stikonas:


bittorf (commented):
this will not work, globbing needs the asterisk outside,
e.g. cp "$foo"* ...

stikonas:
> this will not work, globbing needs the asterisk outside,
> e.g. cp "$foo"* ...

Hmm, yes, it would be better to have * outside. Otherwise it will only work with one file (although that might usually be sufficient but still...).

PR #12, opened 2021-01-15
Add bash 2.05b

fosslinux:
fails at linking


rootfs.sh
```diff
@@ -124,6 +124,10 @@ cp tcc-patched.kaem tmp/after/
 # make 3.80
 get_file https://ftp.gnu.org/gnu/make/make-3.80.tar.gz
 
+# bash 2.05b
+get_file https://ftp.gnu.org/pub/gnu/bash/bash-2.05b.tar.gz
+cp bash-2.05b-makefile tmp/after/
```
stikonas, line 129: Maybe name it `bash-2.05b.mk`. Then get_file will deal with it automatically.

PR #15, opened 2021-01-17
Grep 2.4

stikonas:


sysa/grep-2.4.kaem
```diff
@@ -0,0 +1,13 @@
...
+
+cp ../grep-2.4.mk Makefile
+/after/bin/make
+
+# Install
+/after/bin/cp grep /after/bin/grep
+/after/bin/chmod 755 /after/bin/grep
+
+# Test
+/after/bin/grep --version
```
fosslinux, line 13: Don't need `/after/bin` here

fosslinux (changes requested):

sysa/grep-2.4.kaem
stikonas: Fixed and tested

PR #19, opened 2021-01-21
Heirloom-devutils: yacc and lex

stikonas:
Heirloom `yacc` and `lex` will be needed to bootstrap `flex` and `bison`.

### DIFF
--- a/sysa/heirloom-devtools-070527/heirloom-devtools-070527.kaem
+++ b/sysa/heirloom-devtools-070527/heirloom-devtools-070527.kaem
@@ -1,0 +1,26 @@
+#!/bin/sh
+
+set -ex
+
+mkdir build
+cd build
+
+bindir=/after/bin
+
+# Extract
+bunzip2 ../src/${pkg}.tar.bz2
+tar xf ../src/${pkg}.tar ${pkg}/yacc ${pkg}/lex
+cd ${pkg}
+
+# Prepare and patch
+patch -Np0 -i ../../patches/yacc_remove_wchar.patch
+
+# Build yacc
+cd yacc
+make -f Makefile.mk CC=tcc AR=tcc\ -ar CFLAGS=-DMAXPATHLEN=100\ -DEILSEQ=84\ -DMB_LEN_MAX=100 LDFLAGS=-lgetopt RANLIB=true
+
+# Install yacc
+install yacc ${bindir}
+install -m 644 yaccpar /
+
+cd ../../..
--- a/sysa/heirloom-devtools-070527/heirloom-devtools-070527.kaem
+++ b/sysa/heirloom-devtools-070527/heirloom-devtools-070527.kaem
@@ -6,6 +6,8 @@ mkdir build
 cd build
 
 bindir=/after/bin
+libdir=/after/lib
+lexdir=/lex
 
 # Extract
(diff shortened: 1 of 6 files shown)


sysa/heirloom-devtools-070527/heirloom-devtools-070527.kaem
```diff
@@ -0,0 +1,20 @@
...
+
+# Part 1: Heirloom yacc
+cd ${pkg}/yacc
+
+# Prepare and patch
+patch -Np0 -i ../../../patches/yacc-wchar.patch
+patch -Np0 -i ../../../patches/fakewchar.patch
+
+# Build
+make -f Makefile.mk CC=tcc AR=tcc\ -ar CFLAGS+=-I.\ -DMAXPATHLEN=100\ -DEILSEQ=84\ -DMB_LEN_MAX=100\ -Dswprintf=sprintf\ -Dwcslen=strlen\ -Diswdigit=isdigit\ -Diswxdigit=isxdigit\ -Diswlower=islower\ -Diswupper=isupper\ -Diswspace=isspace\ -Dwcscmp=strcmp\ -Dtowlower=tolower\ -Diswalpha=isalpha\ -Diswalnum=isalnum\ -Dungetwc=ungetc\ -Dwcscpy=strcpy\ -Dgetwc=getc\ -Dputwc=putc LDFLAGS=-lgetopt RANLIB=true
```
fosslinux, line 20: Split on `\`?

fosslinux (changes requested):
A readme bit would also be neat.

sysa/heirloom-devtools-070527/heirloom-devtools-070527.kaem
stikonas: Ok, I can split.

As for `README`, I thought we should finish at least both heirloom `lex` and `yacc` and then describe both in one paragraph.

If you want I can write now, but I thought it's better to wait until we are done with `lex`.

stikonas: Hmm, splitting lines in kaem is a bit tricky. I tried to add `\` at the end of the line and no spaces on the next line but it still didn't work...

fosslinux: Lol forgot it was kaem. Please just remove the `\` then.

fosslinux:
regarding README, yeah, we can just describe both in one para later.

stikonas:
Ok, I think `yacc` hits some mes libc bug because when building on normal system with the same version of tinycc but with glibc it works.

sysa/heirloom-devtools-070527/heirloom-devtools-070527.kaem
stikonas: Agreed on IRC to keep it as it is.

stikonas:
Added lex too although at the moment it fails to compile.

Generated parse.c file contains tokens like

`# define C 257`
instead of
`# define CHAR 257`

and also
```
-static YYCONST yytabelem yyact[]={
+static YYCONST yytabelem y[]={
```

stikonas:
`yacc` is ready to go. Lex will be done in a follow-on PR.

stikonas:
Lex is now working too.

PR #24, opened 2021-01-29
Ls

pder:


sysa/coreutils-5.0/coreutils-5.0.kaem
```diff
@@ -18,6 +18,7 @@ catm config.h
 
 patch -Np0 -i ../../patches/modechange.patch
 patch -Np0 -i ../../patches/mbstate.patch
+patch -Np1 -i ../../patches/ls-strcmp.patch
```
stikonas, line 21: @fosslinux will ask you to switch to `-p0` patch... So you can as well do it now.

Or alternatively, we should ask @fosslinux to change policy on patches. I personally find making `-p1` patches is a bit easier.

sysa/tcc-0.9.27/mes-libc-patched.kaem
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
+
+set -ex
+
+cd src/mes-libc
+
+# Patch
+patch -Np1 -i ../../patches/mes-libc-qsort.patch
```
stikonas, line 8: same here

stikonas (changes requested):

sysa/coreutils-5.0/coreutils-5.0.kaem
fosslinux: I'll consider `-p1`, but let's keep it consistent for now.

stikonas: For now my bash build harness also assumes `-p0`. Although this particular patch is of course before bash. That might need further thought. Especially if we start consuming outside patches instead of making our own.

pder:
Updated the patches to -p0

PR #27, opened 2021-02-03
Add patch headers

fosslinux:
Also cleanup the flex patch a bit

@stikonas  I need you to check if the reasons for all these patches are correct.


sysa/coreutils-5.0/patches/mbstate.patch
```diff
@@ -1,3 +1,6 @@
+mbstat_t is a struct that is required. However, it is not defined by mes libc.
+This implementation was taken from musl.
```
fosslinux, line 2: Is this true @stikonas

sysa/tcc-0.9.27/patches/mes-libc-qsort.patch
```diff
@@ -1,3 +1,6 @@
+Reimplement qswap in a more correct manner that works much more cleanly and
+with larger string sizes.
```
fosslinux, line 2: @pder any other comments

sysa/coreutils-5.0/patches/mbstate.patch
stikonas: No. I think I just took it from my system, so must be glibc 2.32.

stikonas (changes requested):
`scan_l.patch` is essential (despite it only doing cosmetical changes).

`flex-tmp` (which uses lexer generated by `lex`) is not able to process unpatched file.
So we need to (cosmetically) patch scan.l first and compile new `flex`. That `flex` is then good enough and can self-host. Although, we don't rebuild it here since we need to build newer version anyway.

stikonas (changes requested):
`scan_l.patch` is essential (despite it only doing cosmetical changes).

`flex-tmp` (which uses lexer generated by `lex`) is not able to process unpatched file.
So we need to (cosmetically) patch scan.l first and compile new `flex`. That `flex` is then good enough and can self-host. Although, we don't rebuild it here since we need to build newer version anyway.

fosslinux:
Huh, you seem to be right; that is quite odd, but that's ok.

PR #30, opened 2021-02-05
add download cache

bittorf:
overload 'wget' with a function,
so all wget downloads are cached in
ether '/tmp/cachedir-live-bootstrap' or
ENV defined using $CACHEDIR


rootfs.sh
```diff
@@ -14,6 +15,48 @@ sudo mount -t tmpfs -o size=8G tmpfs tmp
 
 LOGFILE="$PWD/tmp/bootstrap.log"
 
+wget()
+{
+	local url="$1"				# e.g. https://server/foo/bar.tgz
+	local dir="$GITDIR/sources/dl"
```
fosslinux, line 21: No need for the `/dl` IMO..

rootfs.sh
```diff
@@ -14,6 +15,48 @@ sudo mount -t tmpfs -o size=8G tmpfs tmp
 
 LOGFILE="$PWD/tmp/bootstrap.log"
 
+wget()
+{
+	local url="$1"				# e.g. https://server/foo/bar.tgz
```
fosslinux, line 20: I really don't like all of these `# e.g.` comments, something that is unclear should be clearly explained in a comment above.

rootfs.sh
```diff
@@ -14,6 +15,48 @@ sudo mount -t tmpfs -o size=8G tmpfs tmp
...
+	local dir="$GITDIR/sources/dl"
+	local file
+
+	test -d "$CACHEDIR" && dir="$CACHEDIR"
```
fosslinux, line 24: instead of doing this you can just set dir to `${CACHEDIR:-${GITDIR}/sources}`

rootfs.sh
```diff
@@ -14,6 +15,48 @@ sudo mount -t tmpfs -o size=8G tmpfs tmp
...
+	file="$( basename "$url" )"		# e.g. bar.tgz
```
fosslinux, line 25: remove spaces around the basename call

rootfs.sh
```diff
@@ -14,6 +15,48 @@ sudo mount -t tmpfs -o size=8G tmpfs tmp
...
+
+	if [ -s "$dir/$file" ]; then
+		cp -v "$dir/$file" .
+	else
+		mkdir -p "$dir"
```
fosslinux, line 30: This can be moved outside of the if/else block.

rootfs.sh
```diff
@@ -14,6 +15,48 @@ sudo mount -t tmpfs -o size=8G tmpfs tmp
...
+		if command wget -O "$dir/$file" "$url"; then
+			cp -v "$dir/$file" .
+		else
+			rm "$dir/$file"
+		fi
+	fi
+
+	test $? -eq 0 && checksum_do "$dir" "$file"
```
fosslinux, line 38: If the file already exists (line 28) I would want the checksumming to happen, and then if the checksumming fails, remove the file and attempt a redownload. This allows us to fix partial downloads that were CTRL-C'd. Then the entire checksumming process should be redone after as it is here.

fosslinux, line 38: No need to test that `$? -eq 0`, we are running with `-e`, please unconditionally checksum.

rootfs.sh
```diff
@@ -3,6 +3,7 @@ set -ex
 
 QEMU_CMD="${1:-qemu-system-x86_64}"	# or 'chroot' or 'minikernel'
 QEMU_RAM="${2:-8G}"
+export GITDIR="$PWD"
```
fosslinux, line 6: On top of this, it would be good to check that we really are running rootfs.sh from the git dir. A quick test that "${GITDIR}/rootfs.sh" exists is not a bad idea, and error out with "rootfs.sh must be run from the base of the repo".

rootfs.sh
```diff
@@ -14,6 +15,48 @@ sudo mount -t tmpfs -o size=8G tmpfs tmp
...
+			cp -v "$dir/$file" .
+		else
+			rm "$dir/$file"
+		fi
+	fi
+
+	test $? -eq 0 && checksum_do "$dir" "$file"
+}
+
+checksum_do()
```
fosslinux, line 41: A much simpler implementation of this function could:

set file to the file to be checked (no need for dir)
set checksum_file (or store or whatever) to the path to the sha256sum file
something like:
```
# get the appropriate line from the checksum file, but..
if ! checksum="$(grep "$(basename ${file}")"; then
    # if it is not there, error out
    echo "There is no checksum for ${file}! Please fix."
fi

# Now check the checksum
pushd "$(dirname ${file})"
echo "${checksum}" | sha256sum -c
popd
```

Much easier to read and less complex.

fosslinux (changes requested):
Please also rebase.

rootfs.sh
stikonas: You don't need outer quotes either. `file=$(basename "$url")` will work.

rootfs.sh
stikonas: maybe `basename "$0"` would be better instead of $PWD?

rootfs.sh
bittorf: ok,  simplied the thing as much as possible, rebased and add explanations to commit-message.

rootfs.sh
```diff
@@ -14,6 +20,44 @@ sudo mount -t tmpfs -o size=8G tmpfs tmp
 
 LOGFILE="$PWD/tmp/bootstrap.log"
 
+wget()
+{
+	local url="$1"
+	local dir="${CACHEDIR:-$GITDIR/sources/dl}"
```
fosslinux, line 26: `/dl` is still unresolved, do you forsee a particular usecase for sources outside of tarballs?

rootfs.sh
fosslinux: Please act on this

rootfs.sh
fosslinux: This doesn't even work now, checksum is not calculated. Line 53 is redundant, it does nothing.

fosslinux (changes requested):

rootfs.sh
fosslinux: This doesn't even work now, checksum is not calculated. Line 53 is redundant, it does nothing.

rootfs.sh
fosslinux: @stikonas we cannot assume that currently, other parts of the script expect to be run from the root of the repository

rootfs.sh
```diff
@@ -4,6 +4,12 @@ set -ex
 QEMU_CMD="${1:-qemu-system-x86_64}"	# or 'chroot' or 'minikernel'
 QEMU_RAM="${2:-8G}"
 
+GITDIR="$PWD/$(dirname "$0")"
+test -f "$GITDIR/rootfs.sh" || {
+	echo 'must be run from base of repo'
+	false
+}
```
fosslinux, line 11: This is a poor style, use an if block instead

```
if [ -f "$GITDIR/rootfs.sh" ]; then
    echo "rootfs must be run from the base of the repo
    exit 1
fi
(false is also incorrect here, exit 1 should be used instead, as that doesn't exploit the use of set -e)

File: rootfs.sh (hunk 5 above)
Comment (bittorf) [reply rootfs.sh:34]: the mkdir? but is is only *needed* in the else-case - or do you mean something different?

File: rootfs.sh (hunk 8 above)
Comment (bittorf) [reply rootfs.sh:45]: it works here and in CI. just try to *change* a checksum in SHA256-file or the remove it. Line 53 is for getting the var "checksum" out of the line (firstword in line).

File: rootfs.sh (hunk 9 above)
Comment (bittorf) [reply rootfs.sh:26]: the usecase is in the commit-message:
```
downloads can be made to every location, not
necessary to 'sources', hence to special dir 'dl'.
```

File: rootfs.sh (hunk 10 above)
Comment (bittorf) [reply rootfs.sh:11]: if fixed it:
we have `set -e`, so when i try to run it from outside, it stop the script:

```
bastian@ryzen:~/software/live-bootstrap-bittorf$ live-bootstrap/rootfs.sh 
+ QEMU_CMD=qemu-system-x86_64
+ QEMU_RAM=8G
++ dirname live-bootstrap/rootfs.sh
+ GITDIR=/home/bastian/software/live-bootstrap-bittorf/live-bootstrap
+ test -f ./rootfs.sh
+ echo 'must be run from base of repo'
must be run from base of repo
+ false
```

File: rootfs.sh (hunk 5 above)
Comment (fosslinux) [reply rootfs.sh]: oh, true

File: rootfs.sh (hunk 8 above)
Comment (fosslinux) [reply rootfs.sh:45]: OHH, I get it now. A for loop is quite a bad way to do this, as for is made for looping, and splits on spaces and stuff too. Since a sha256sum file is tab delimited, a much better way to do that construct would be just (replacing lines 52 and 53): `checksum="$(sha256sum <"$dir/$file" | cut -f1)"`, cut takes the first field tab delimited.

File: rootfs.sh [hunk 11]
```diff
@@ -14,6 +20,44 @@ sudo mount -t tmpfs -o size=8G tmpfs tmp
...
+{
+	local dir="$1"
+	local file="$2"
+	local checksum out
+	local store="$GITDIR/SHA256SUMS.sources"
+
+	out="$( sha256sum <"$dir/$file" )"
+	for checksum in $out; do break; done
+
+	grep -q ^"$checksum  $file"$ "$store" || {
```
stikonas, line 55: It's probably a minor thing but I would feel more comfortable if we rely on sha256sum to check whether hash matches.
I.e. instead of running sha256sum and then use grep to compare, run grep to get sha256sum and run sha256sum --check to verify.

Something like
```
if checksum=$(grep $file SHA256SUMS.sources); then
    (cd $dir; echo $checksum | sha256sum --check )
else
    echo "error"
fi
```
This also has the advantage of printing `tar-1.12.tar.gz: OK` to log.

fosslinux: Yeah, I agree.

rootfs.sh
```diff
@@ -14,6 +20,35 @@ sudo mount -t tmpfs -o size=8G tmpfs tmp
...
+{
+	local dir="$1"
+	local file="$2"
+	local line
+	local store="$GITDIR/SHA256SUMS.sources"
+
+	if line=$(grep "[[:space:]][[:space:]]$file"$ "$store"); then
+		(cd "$dir" && echo "$line" | sha256sum -c)
+	else
+		echo "# error: add hash to '$store' or remove e.g. partitial downloaded file"
```
fosslinux, line 47: Typo here, but also not an amazing message, a much better message would be "Checksum mismatch or not found!"

rootfs.sh
```diff
@@ -4,6 +4,12 @@ set -ex
 QEMU_CMD="${1:-qemu-system-x86_64}"	# or 'chroot' or 'minikernel'
 QEMU_RAM="${2:-8G}"
 
+GITDIR="$PWD/$(dirname "$0")"
+test -f rootfs.sh || {
+	echo 'must be run from base of repo'
+	false
+}
```
fosslinux, line 11: Please change to if block and use exit 1.

rootfs.sh
```diff
@@ -14,6 +20,35 @@ sudo mount -t tmpfs -o size=8G tmpfs tmp
...
+	local dir="$1"
+	local file="$2"
+	local line
+	local store="$GITDIR/SHA256SUMS.sources"
+
+	if line=$(grep "[[:space:]][[:space:]]$file"$ "$store"); then
+		(cd "$dir" && echo "$line" | sha256sum -c)
+	else
+		echo "# error: add hash to '$store' or remove e.g. partitial downloaded file"
+		false
```
fosslinux, line 48: Please use exit 1

fosslinux (changes requested):

fosslinux (approved):

PR #36, opened 2021-02-10
Add bison 3.4.1.

stikonas:


sysa/bison-3.4.1/patches/scan-skel-missing-include.patch
```diff
@@ -0,0 +1,30 @@
+# SPDX-FileCopyrightText: 2020 Giovanni Mascellani <gio@debian.org>
```
fosslinux, line 1: The two missing include patches can probably be merged.

Also our patches don't generally have `# ` at the beginning of the SPDX lines (as per all the other patches).

fosslinux (changes requested):

stikonas:
Minor update: added `yacc` symlink.

fosslinux (approved):

PR #39, opened 2021-02-14
Checksumming

fosslinux:
TODO: Reuse


fosslinux:
This should now be finished and ready

rootfs.sh
```diff
@@ -126,11 +122,12 @@ tar -C tmp/after/tar-1.12/src -xf "tmp/after/tar-1.12/src/$(basename $url)" --st
 get_file() {
     url=$1
     make_build=${2:-0}
+    output_filename=$3
     ext="${url##*.}"
     if [ "$ext" = "tar" ]; then
-        bname=$(basename "$url" ".tar")
+        bname="$(basename "${output_filename:-${url}}" ".tar")"
```
stikonas, line 128: External quotes are not necessary here since expansion does not happen when assigning to another variable.

stikonas (commented):
It looks good, although, it might be hard to manually update all hashes if we change some core component... E.g. patch tcc which will affect all hashes.

rootfs.sh
fosslinux: :+1:

fosslinux:
I don't believe the tcc use case will happen very often, however, I do understand the issue. I'm considering some solutions to this including:

- I want to split out all of the rms we do in `src_install()` anyways into a `src_clean()`, so perhaps we can add some option to disable that.
- We could also add a way of making the resulting binaries be sha256summed automatically after something is set.

However, I'll add that after this PR.

I'm just adding Perl now then I'll merge

PR #43, opened 2021-02-18
Minor cleanup

stikonas:


README.md
```diff
@@ -357,5 +357,5 @@ We now use `perl` from the previous stage to recreate pre-generated files that
 are shipped in perl 5.003. But for now we still need to use handwritten makefile
 instead of `./Configure` script.
 
-### Part 31: perl 5.004_05
+#### Part 31: perl 5.004_05
```
fosslinux, line 360: Newline here while we are at it?

bittorf: can you please also these:
```
bastian@ryzen:~/live-bootstrap$ codespell | grep -v " defin " | grep -v " pres " | grep -v " toke "
./DEVEL.md:44: Permissable ==> Permissible
./DEVEL.md:50: nonexistant ==> nonexistent
./sysa/bison-3.4.1/files/parse-gram.c:50: boostrapping ==> bootstrapping
./sysa/bison-3.4.1/files/parse-gram.y:491: reudce ==> reduce
./sysa/heirloom-devtools-070527/patches/lex_remove_wchar.patch:11: Similarily ==> Similarly
./sysa/mescc-tools-extra/src/cp.c:113: dosen't ==> doesn't
./sysa/mescc-tools-extra/src/cp.c:313: initalization ==> initialization
./sysa/musl-1.1.24/patches/fenv.patch:7: migt ==> might, midget
```

stikonas: Some of these are upstream files (bison and probably mescc-tools-extra?). It would be better to report those upstream first. I'll change the rest.

bittorf (commented):
please better use:
```
test -f /dev/null && rm /dev/null
```
on systems with a real `/dev/null` this throws an error.

fosslinux:
There is no real `/dev/null` though? We are operating in live-bootstrap.

stikonas:
> There is no real `/dev/null` though? We are operating in live-bootstrap.
> 
> Please rebase.

Indeed there is no real `/dev/null` yet. even if we create a real /dev/null in live-bootstrap, for some reason musl's configure script somehow replaces it with a file containing contents that were sent to `/dev/null`.

We can create real `/dev/null` later (probably before autotools) with `mknod`.

PR #45, opened 2021-02-19
Rebuild musl libc after building tcc-musl

pder:
This fixes an issue with printf not printing doubles or floats
correctly, and probably other subtle issues.


stikonas (approved):
Should we add a new "Part" for musl recompilation?

fosslinux:
Yes, I think we should.

stikonas:
Also as discussed on IRC, it probably makes sense to rebuild tcc again, so that tcc itself has working floats (in case it uses them).

fosslinux:
Please rebase. It would be nice to also rebuild tcc as @stikonas mentioned about. I can do checksums if you want cause this is rather painful.

fosslinux:
Merged with a couple of changes.

PR #48, opened 2021-02-24
Add Perl 5.6.2.

stikonas:
Unfortunately, this does not yet unlock any newer autoconf/automake versions (those seem to depend on full perl, but maybe we can add a few more steps to makefile later and get support extensions.

With latest fix, perl 5.18 also seems reachable.


stikonas:
Also added `define HAS_READDIR`. That might help with some perl scripts (in particular might be needed if we want to build full perl)

stikonas:
(just rebased on top of merged `bzip2` PR)

fosslinux (approved):

PR #52, opened 2021-02-26
Autotools

stikonas:
Ignore the first commit, it's part of another PR.

* First we populate device nodes in `/dev`.
* Then manually build and install `autoconf 2.52`.
* Use `autoconf-2.52` to simplify build of `automake`.
* Rebuild `automake` with `autoconf` and `automake`
* Rebuild `autoconf-2.52`
* Build `autoconf-2.13`.


sysa/autoconf-2.52/autoconf-2.52.sh
```diff
@@ -0,0 +1,36 @@
...
+populate_device_nodes() {
+    # http://www.linuxfromscratch.org/lfs/view/6.1/chapter06/devices.html
+    test -c /dev/console || mknod -m 622 /dev/console c 5 1
+    test -c /dev/null || mknod -m 666 /dev/null c 1 3
+    test -c /dev/zero || mknod -m 666 /dev/zero c 1 5
+    test -c /dev/ptmx || mknod -m 666 /dev/ptmx c 5 2
+    test -c /dev/tty || mknod -m 666 /dev/tty c 5 0
+    test -c /dev/random || mknod -m 444 /dev/random c 1 8
+    test -c /dev/urandom || mknod -m 444 /dev/urandom c 1 9
+}
```
fosslinux, line 27: IMO, this belongs in a separate stage. It is not completely related to autoconf and we definitely will use it for other things.

stikonas: do you think I can put this function to helpers.sh?

fosslinux: Dislike that, it's not really a helper; I think it would be ok in run.sh though

sysa/autoconf-2.52/autoconf-2.52.sh
```diff
@@ -0,0 +1,36 @@
...
+
+src_install() {
+    install autoconf ${PREFIX}/bin/
```
fosslinux, line 30: Rename the binary to `autoconf252` - this is the standard practice done in most distributions.

fosslinux (changes requested):

sysa/autoconf-2.52/autoconf-2.52.sh
stikonas: renamed to autoconf-2.52

fosslinux (approved):

PR #56, opened 2021-03-04
bash-3.2.57: Build new package to replace bash 2.05b

pder:
Resulting build now works interactively


parts.rst
```diff
@@ -419,3 +419,12 @@ tcc 0.9.27 (musl v3)
...
 We rebuild tcc against new musl and without a patch to ignore duplicate symbols.
+
+bash 3.2.57
+==========
+
+Up to this point, our build of ``bash`` could run scripts but could not be used
+interactively.  This newer version of ``bash`` compiles without any patches and
+provides new features as well as being able to be used as an interactive shell.
+We use autoconf 2.52 to regenerate the configure script and bison to recreate
+some included generated files.
```
stikonas, line 430: Maybe we can make it a bit clearer that interactivity is due to `readline` and not just newer `bash`?

sysa/run.sh
```diff
@@ -91,4 +91,6 @@ populate_device_nodes
 # Rebuild tcc-musl using new musl
 build tcc-0.9.27 tcc-musl-pass3.sh checksums/tcc-musl-pass3
 
+build bash-3.2.57
+
 echo "Bootstrapping completed."
```
stikonas, line 96: Maybe we can add

`exec bash` at the very end after `echo "Bootstrapping completed."`

stikonas (changes requested):
Can you also update `README.rst` where we mention that we'll soon have interactive bash.

sysa/run.sh
pder: Instead of exec bash which will never exit, I think we talked about doing exec bash run2.sh which starts the next script under this new interpreter.  Also won't exec bash cause our ci to fail?

stikonas: Well, we definitely can do `exec bash run2.sh` too.

I think CI is scanning for `Bootstrapping completed.` Once it is done CI is killed and marked as successfull.
I think it will still work fine with `exec bash` since current CI doesn't exit anyway, the kernel simply panics and qemu is stuck forever. That's why I was thinking that instead of qemu stuck forever, it would be nicer to show terminal.

But maybe then it's better to do
`exec bash run2.sh`
Move `echo "Bootstrapping completed"` to `run2.sh` and then do `exec bash` after that `echo`.

parts.rst
```diff
@@ -419,3 +419,12 @@ tcc 0.9.27 (musl v3)
 ====================
 
 We rebuild tcc against new musl and without a patch to ignore duplicate symbols.
+
+bash 3.2.57
+==========
```
stikonas, line 424: missing one `=`

sysa/run2.sh
```diff
@@ -0,0 +1,3 @@
+echo "Bootstrapping completed."
```
stikonas, line 1: please add shebang and spdx stuff.

sysa/run2.sh
```diff
@@ -0,0 +1,14 @@
...
+# SPDX-FileCopyrightText: 2021 Andrius Štikonas <andrius@stikonas.eu>
+# SPDX-FileCopyrightText: 2021 fosslinux <fosslinux@aussies.space>
+# SPDX-FileCopyrightText: 2021 Paul Dersey <pdersey@gmail.com>
+#
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+set -e
+. helpers.sh
+
+export PREFIX=/after
```
fosslinux, line 12: Is this required? Shouldn't it propagate from the previous script?

fosslinux (approved):

PR #59, opened 2021-03-08
coreutils: Build env.

stikonas:


pder:
Maybe also add dirname since that will also be needed?

PR #61, opened 2021-03-11
Rebuild sed and add libtool

stikonas:
Rebuilds sed against musl


fosslinux:
Freaking finally got back into gh, now i can review.

parts.rst
```diff
@@ -408,6 +419,11 @@ In particular we can now use full featured ``ar`` instead of ``tcc -ar``,
 the GNU linker ``ld``, which allows us building shared libraries,
 and the GNU assembler ``as``.
 
+libtool 1.4
+===========
+
+Rebuild libtool to use GNU binutils.
```
fosslinux, line 425: This is so `libtool` uses GNU binutils, not so that it compiles with it, right? If so, make that clearer.

sysa/sed-4.0.7/sed-4.0.7.sh
```diff
@@ -0,0 +1,21 @@
...
+#
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+src_unpack() {
+    mkdir sed-4.0.7
+    cp -R ../src/* sed-4.0.7/
+    cp ../mk/main.mk sed-4.0.7/Makefile
+    cd sed-4.0.7
+    rm *.o lib/*.o lib/libsed.a
+    cd ..
```
fosslinux, line 11: I'm wondering instead of doing all this dumb cleanup, what if we used http://ftp.gnu.org/gnu/sed/sed-4.0.9.tar.gz or http://ftp.gnu.org/gnu/sed/sed-4.0.7.tar.gz (and extract it to a different dir), or even a newer sed.

parts.rst
```diff
@@ -251,6 +251,11 @@ tcc 0.9.27 (musl v2)
 Now that we have a ‘fixed’ ``musl``, we now recompile ``tcc`` as ``tcc``
 uses floats extensively.
 
+sed 4.0.9
+=========
+
+``sed`` is rebuilt against musl.
```
fosslinux, line 257: A slightly newer `sed` ...?

fosslinux (changes requested):

PR #64, opened 2021-03-14
Add automake 1.5.

stikonas:
P.S. I removed mentions of `miniperl` in `parts.rst` because now we ship some perl modules too. It's probably still easier to bootstrap those particular versions of `automake` and `autoconf` (using `sed`).


fosslinux (approved):

stikonas:
(just rebased on top of master)

PR #68, opened 2021-03-15
Sed after tar and gzip

pder:
This PR moves the sed build to after tar and gzip which allows the sed submodule to be removed and instead be built from a downloaded tarball.


sysa/tar-1.12/files/getdate_stub.c
```diff
@@ -0,0 +1,12 @@
+/*
+ * SPDX-FileCopyrightText: 2021 Paul Dersey <pdersey@gmail.com>
+ *
+ * SPDX-License-Identifier: GPL-3.0-or-later
```
stikonas, line 4: Can we have the same license as tar 1.12?

stikonas (approved):

PR #77, opened 2021-03-25
Xz

stikonas:
* Fix GCC pregen stuff
* xz 5.0.5
* More autotools

PR #80, opened 2021-03-30
Libtool 2.2.4 and fix GCC pregen files

stikonas:
Fixes GCC not to use pre-generated Makefile.in by not using top-level configure.


stikonas:
Alternatively, the whole #77 can be merged

PR #85, opened 2021-04-02
Add support for DESTDIR variable in make install.

stikonas:


sysa/coreutils-5.0/mk/main.mk
```diff
@@ -13,7 +13,8 @@ CC      = tcc
 LD      = tcc
 AR      = tcc -ar
 
-bindir=/after/bin
+PREFIX=/after
+bindir=$(PREFIX)/bin
```
fosslinux, line 17: This should be `$(DESTDIR)$(PREFIX)` also?

stikonas: That's for kaem build.

PR #91, opened 2021-04-07
Port rootfs.sh to Python

stikonas:
Port rootfs to python and add pylint linter.


rootfs.py
```diff
@@ -0,0 +1,79 @@
...
+                        default="qemu-system-x86_64")
+    parser.add_argument("-r", "--qemu-ram", help="Memory (in megabytes) allocated to QEMU VM",
+                        default=8)
+    parser.add_argument("-k", "--kernel", help="Kernel to use (default is ./kernel)",
+                        default="kernel")
+
+    parser.add_argument("-m", "--minikernel", help="Build minikernel",
+                        action="store_true")
+
+
```
fosslinux, line 34: Again double newline.

Also you probably want some kind of option validation here that ensures only one of chroot/qemu/minikernel is selected, otherwise there is a whole host of issues.

rootfs.py
```diff
@@ -0,0 +1,79 @@
...
+        if not args.chroot:
+            make_initramfs(sysa.tmp_dir, initramfs_path)
+
+        bootstrap(args, sysa.tmp_dir, initramfs_path)
+    finally:
+        sysa.unmount_tmpfs()
+
+def make_initramfs(tmp_dir, initramfs_path):
+    file_process = subprocess.Popen(["find", tmp_dir, "-name", "*", '-printf', "%P\n"],
+                                    stdout=subprocess.PIPE)
```
fosslinux, line 50: Not sure how much I like shelling out to find... it shouldn't be too hard to do this in native Python?

rootfs.py
```diff
@@ -0,0 +1,79 @@
...
+def bootstrap(args, tmp_dir, initramfs_path):
+    init = os.path.join(os.sep, 'bootstrap-seeds', 'POSIX', args.arch, 'kaem-optional-seed')
+
+    print("Bootstrapping %s" % (args.arch))
+    if args.chroot:
+        run('sudo', 'env', '-i', 'PATH=/bin', 'chroot', tmp_dir, init)
+        return
+
+    if args.minikernel:
+        run('git', 'clone', 'depth', '1', '--branch', 'v0.4', 'https://github.com/bittorf/kritis-linux.git')
```
fosslinux, line 64: `--depth`

rootfs.py
```diff
@@ -0,0 +1,79 @@
...
+        run('kritis-linux/ci_helper.sh', '--arch', 'x86_64', '--ramsize',
+            '-m', str(args.qemu_ram) + 'G', '--kernel', '5.10.8', '--initrd', initramfs_path)
```
fosslinux, line 66: Parser description says "Memory (in megabytes) allocated to QEMU VM".. but you have G here.

rootfs.py
```diff
@@ -0,0 +1,79 @@
...
+
+def make_initramfs(tmp_dir, initramfs_path):
+    file_process = subprocess.Popen(["find", tmp_dir, "-name", "*", '-printf', "%P\n"],
+                                    stdout=subprocess.PIPE)
+    with open(initramfs_path, "w") as initramfs:
+        run("cpio", "--format", "newc", "--create", "--directory", tmp_dir,
+            stdin=file_process.stdout, stdout=initramfs)
+
+def bootstrap(args, tmp_dir, initramfs_path):
+    init = os.path.join(os.sep, 'bootstrap-seeds', 'POSIX', args.arch, 'kaem-optional-seed')
```
fosslinux, line 56: We only need to eval this for chroot.

rootfs.py
```diff
@@ -0,0 +1,79 @@
...
+import os
+import subprocess
+
+import sysa
+from lib.utils import run
+
+def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument("-a", "--arch", help="Bootstrap architecture",
+                        default="x86")
```
fosslinux, line 16: This needs to have a catch or some other kind of check around it that ensures it only accepts i386/x86, we don't want people getting the idea they can use other architectures.

sysa.py
```diff
@@ -0,0 +1,431 @@
...
+git_dir = os.path.dirname(os.path.join(__file__))
+
+_arch = None
+_preserve_tmp = None
+tmp_dir = None
+sysa_dir = None
+after_dir = None
+
+def create_sysa(arch, preserve_tmp, tmpdir):
+    global _arch, _preserve_tmp, tmp_dir, sysa_dir, after_dir
```
fosslinux, line 24: Python globals persist outside files, no? That would be a problem for sysb and higher.

sysa.py
```diff
@@ -0,0 +1,431 @@
...
+            raise Exception("Checksum mismatch")
+
+    raise Exception("File checksum is not yet recorded")
+
+def download_file(url, file_name=None):
+    cache_dir = os.path.join(git_dir, 'sources')
+
+    if file_name is None:
+        file_name = os.path.basename(url)
+    abs_file_name = os.path.join(cache_dir, file_name)
```
fosslinux, line 66: Do we need a separate variable here?

sysa.py
```diff
@@ -0,0 +1,431 @@
...
+    # Determine installation directory
+    target_name = _get_target(file_name)
+    target_src_dir = os.path.join(after_dir, target_name, 'src')
+
+    # Install base files
+    src_tree = os.path.join(sysa_dir, target_name)
+    copytree(src_tree, after_dir)
+    os.mkdir(target_src_dir)
+
+    for i,_ in enumerate(urls):
```
fosslinux, line 126: `i, _` preferably

fosslinux, line 126: `i, _` preferably

rootfs.py
```diff
@@ -0,0 +1,79 @@
...
+
+    # QEMU arguments
+    parser.add_argument("-q", "--qemu-cmd", help="QEMU command",
+                        default="qemu-system-x86_64")
+    parser.add_argument("-r", "--qemu-ram", help="Memory (in megabytes) allocated to QEMU VM",
+                        default=8)
+    parser.add_argument("-k", "--kernel", help="Kernel to use (default is ./kernel)",
+                        default="kernel")
+
+    parser.add_argument("-m", "--minikernel", help="Build minikernel",
```
fosslinux, line 31: Use minikernel maybe? Build minikernel kinda implies it's just being buildt and not used, or build inside live-bootstrap.

rootfs.py
```diff
@@ -0,0 +1,79 @@
...
+from lib.utils import run
+
+def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument("-a", "--arch", help="Bootstrap architecture",
+                        default="x86")
+    parser.add_argument("-c", "--chroot", help="Run inside chroot",
+                        action="store_true")
+    parser.add_argument("-p", "--preserve", help="Do not unmount temporary dir",
+                        action="store_true")
```
fosslinux, line 20: This is probably clearer with a default.

rootfs.py
```diff
@@ -0,0 +1,79 @@
...
+    parser.add_argument("-t", "--tmpdir", help="SysA temporary directory")
```
fosslinux, line 21: Can we make the description Temporary directory or otherwise distinguish the argument for sysb etc?

fosslinux (changes requested):
Reasonably good.

I think that I prefer sysa as a class more. It seems pretty clear usecase for OOP and the globals seem really hacky. I think it is easier to reason about with a class, but up to you.

There should be many more comments. This needs to be easily reproducible by someone without using this script, arguably the most important part of live-bootstrap that needs to be manually doable. If the function descriptions, particularly around get_file and the like, and the comments scattered around there, are replicated in other functions, that would be great.

rootfs.py
stikonas: I've now replaced it with `Temporay directory`. I don't think we'll need a separte temp dir for sysb as we have to create it inside sysa.

rootfs.py
stikonas: Can you clarify this? store_true automatically has default value of `False`. If `-p` is specified than `args.preserve=True`.

stikonas:
CI failed due to my python being newer. I guess need to fix code a bit to work with older python...

sysa.py
stikonas: Hmm, I think I used a separate variable not to accidentally modify file_name that is passed from outside into the function.

fosslinux (changes requested):
The python script looks great. Can we perhaps shorten or remove `.pylintrc`? It's very long and I'm not sure if its really necessary to check the entire thing into the repo.

Issue #99, opened 2021-04-14
bash-5.1: frequently 50% ABORT: bad checksum


bittorf:
seen with 2423b77114221f515500f6345b465d74db7688ed and qemu-386 with kernel 3.18.140 and 3G RAM.
a workaround is the add an additional sleep in `sysa/bash-5.1/bash-5.1.sh` like:

```
- sed -i 's/sleep 3/sleep 3; sync/' builtins/psize.sh
+ sed -i 's/sleep 3/sleep 3; sync; sleep 3/' builtins/psize.sh
```

see:
http://intercity-vpn.de/bootstrap/lbmulti-log-1617621903.txt

```
[...]
01h31m48s | gcc -L./builtins -L./lib/readline -L./lib/readline -L./lib/glob -L./lib/tilde  -L./lib/sh  -L./lib/termcap -rdynamic -g -O2 -Wno-paren
theses -Wno-format-security  -static -o bash shell.o eval.o y.tab.o general.o make_cmd.o print_cmd.o  dispose_cmd.o execute_cmd.o variables.o copy
_cmd.o error.o expr.o flags.o jobs.o subst.o hashcmd.o hashlib.o mailcheck.o trap.o input.o unwind_prot.o pathexp.o sig.o test.o version.o alias.o
 array.o arrayfunc.o assoc.o braces.o bracecomp.o bashhist.o bashline.o  list.o stringlib.o locale.o findcmd.o redir.o pcomplete.o pcomplib.o synt
ax.o xmalloc.o  -lbuiltins -lglob -lsh -lreadline -lhistory ./lib/termcap/libtermcap.a -ltilde
01h31m49s | ls -l bash
01h31m49s | -rwxr-xr-x    1 0        0         2621878 Apr  5 12:55 bash
01h31m49s | size bash
01h31m49s | text           data     bss     dec     hex filename
01h31m49s | 1043217       20000   45968 1109185  10ecc1 bash
01h31m49s | make[1]: Entering directory `/after/bash-5.1/build/bash-5.1/support'
01h31m49s | rm -f man2html.o
01h31m49s | gcc -c  -DHAVE_CONFIG_H -DSHELL  -I/after/bash-5.1/build/bash-5.1 -I..   -g -O2 -Wno-parentheses -Wno-format-security man2html.c
[...]

PR #104, opened 2021-04-15
fixes bash build.

bittorf:
fixes: https://github.com/fosslinux/live-bootstrap/issues/99

i my usecase (qemu, kernel 3.18.140) about 50% of all
builds are failing with a wrong binary hash.
the culprit is a wrong generation of 'pipesize.h' during:

```
/bin/sh ./psize.sh > pipesize.h
```

normally generated 'pipesize.h' looks like:
```
/*
* pipesize.h
*
* This file is automatically generated by psize.sh
* Do not edit!
*/

#define PIPESIZE 65536
```

in the error case the last line has 512 instead of 65536.

### DIFF
--- a/sysa/bash-5.1/bash-5.1.sh
+++ b/sysa/bash-5.1/bash-5.1.sh
@@ -1,5 +1,6 @@
 # SPDX-FileCopyrightText: 2021 Paul Dersey <pdersey@gmail.com>
 # SPDX-FileCopyrightText: 2021 Andrius Štikonas <andrius@stikonas.eu>
+# SPDX-FileCopyrightText: 2021 Bastian Bittorf <bb@npl.de>
 #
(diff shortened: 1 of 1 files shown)


stikonas:
Looks fine to me although, I'm not convinced that patch makes it more readable here. Maybe shipping our own new version of `builtins/psize.sh` in `files/` dir is more readable?

stikonas (approved):

fosslinux:
Oh, I didn't realise that it was a whole file change.. .shipping it is better.

At this point, why don't we just ship psize.h, the script is literally echoing stuff..

bittorf:
I changed the oneliner again  
to just overwrite `builtins/psize.sh`.

see:
https://github.com/fosslinux/live-bootstrap/pull/104/commits/50417514de5027c721314e47f613356ad3f20db4

PR #105, opened 2021-04-16
rootfs.py: minikernel:

bittorf:
use branch v0.7 from kritis-linux and change settings
to be more i386-alike and add essential kernel feature
'flock' using latest kernel 3.18.140 from 3.18.x-branch

### DIFF
--- a/rootfs.py
+++ b/rootfs.py
@@ -8,11 +8,13 @@ you can run bootstap inside chroot.
 
 # SPDX-License-Identifier: GPL-3.0-or-later
 # SPDX-FileCopyrightText: 2021 Andrius Štikonas <andrius@stikonas.eu>
+# SPDX-FileCopyrightText: 2021 Bastian Bittorf <bb@npl.de>
 
 import argparse
 import glob
 import os
 import subprocess
+import shutil
 
 from sysa import SysA
 from lib.utils import run
@@ -72,7 +74,7 @@ def make_initramfs(tmp_dir, initramfs_path):
 
     with open(initramfs_path, "w") as initramfs:
         cpio = subprocess.Popen(["cpio", "--format", "newc", "--create", "--directory", tmp_dir],
-                             stdin=subprocess.PIPE, stdout=initramfs)
+                                stdin=subprocess.PIPE, stdout=initramfs)
         cpio.communicate(input='\n'.join(file_list).encode())
 
 def bootstrap(args, tmp_dir, initramfs_path):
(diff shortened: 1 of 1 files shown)


rootfs.py
```diff
@@ -84,10 +85,19 @@ def bootstrap(args, tmp_dir, initramfs_path):
...
-            'https://github.com/bittorf/kritis-linux.git')
-        run('kritis-linux/ci_helper.sh', '--arch', 'x86_64', '--ramsize',
-            '-m', str(args.qemu_ram) + 'M', '--kernel', '5.10.8', '--initrd', initramfs_path)
+        os.system("test -d kritis-linux && rm -fR kritis-linux")
+        os.system("git clone --depth 1 --branch v0.7 https://github.com/bittorf/kritis-linux.git")
+        run('kritis-linux/ci_helper.sh',
+                '--private',
+                '--multi',  '1',
+                '--repeat', '1',
+                '--arch', 'i386',
```
stikonas, line 94: Is minikernel going to stay `i386` only? Or would other arches eventually work too? If other arches are planned, maybe instead of hardcoding `i386` we can use args.arch (note that it is `args.arch=x86` and not `i386`, so we might need some map)

bittorf: you are right, it support just all archs. the map is internally handled already, but let us focus now on i386

rootfs.py
```diff
@@ -84,10 +85,19 @@ def bootstrap(args, tmp_dir, initramfs_path):
         return
 
     if args.minikernel:
-        run('git', 'clone', '--depth', '1', '--branch', 'v0.4',
-            'https://github.com/bittorf/kritis-linux.git')
-        run('kritis-linux/ci_helper.sh', '--arch', 'x86_64', '--ramsize',
-            '-m', str(args.qemu_ram) + 'M', '--kernel', '5.10.8', '--initrd', initramfs_path)
+        os.system("test -d kritis-linux && rm -fR kritis-linux")
```
fosslinux, line 88: Please use a python construct.

rootfs.py
```diff
@@ -84,10 +85,19 @@ def bootstrap(args, tmp_dir, initramfs_path):
...
+        os.system("git clone --depth 1 --branch v0.7 https://github.com/bittorf/kritis-linux.git")
```
fosslinux, line 89: Use our wrapper function run()

fosslinux (changes requested):

rootfs.py
bittorf: ok, done

rootfs.py
bittorf: ok, done

rootfs.py
```diff
@@ -84,10 +86,21 @@ def bootstrap(args, tmp_dir, initramfs_path):
...
 
     if args.minikernel:
-        run('git', 'clone', '--depth', '1', '--branch', 'v0.4',
-            'https://github.com/bittorf/kritis-linux.git')
-        run('kritis-linux/ci_helper.sh', '--arch', 'x86_64', '--ramsize',
-            '-m', str(args.qemu_ram) + 'M', '--kernel', '5.10.8', '--initrd', initramfs_path)
+        if path.isdir('kritis-linux'):
+            shutil.rmtree('kritis-linux')
+        
+        run('git clone', '--depth 1', '--branch', 'v0.7', 'https://github.com/bittorf/kritis-linux.git')
```
stikonas, line 92: 'git', 'clone'

rootfs.py
```diff
@@ -84,10 +86,21 @@ def bootstrap(args, tmp_dir, initramfs_path):
         return
 
     if args.minikernel:
-        run('git', 'clone', '--depth', '1', '--branch', 'v0.4',
-            'https://github.com/bittorf/kritis-linux.git')
-        run('kritis-linux/ci_helper.sh', '--arch', 'x86_64', '--ramsize',
-            '-m', str(args.qemu_ram) + 'M', '--kernel', '5.10.8', '--initrd', initramfs_path)
+        if path.isdir('kritis-linux'):
```
stikonas, line 89: `os.path`

rootfs.py
bittorf: ok, done

rootfs.py
bittorf: ok, done

stikonas (approved):

Issue #93, opened 2021-04-08
automake-1.13.4: infrequently 20% ABORT: Argument "" isn't numeric in bitwise and (&) at ./gen-testsuite-part line 67.


bittorf:
seen with dcec416 and qemu-386 with kernel 3.18.140 and 3G RAM
happens in about 20% (24 testruns)

bad: http://intercity-vpn.de/bootstrap/bootstrap.log-multilog-1-1617888807.txt
good: http://intercity-vpn.de/bootstrap/bootstrap.log-multilog-4-1617888807.txt

```
01h12m46s | automake-1.13.4: compiling source.
01h12m46s | GEN      t/testsuite-part.am
01h12m47s | Argument "" isn't numeric in bitwise and (&) at ./gen-testsuite-part line 67.
01h12m47s | make: *** [t/testsuite-part.am] Error 2
01h12m47s | Subprocess error 512
01h12m47s | ABORTING HARD
```

bittorf:
now that other issues are fixed, 30% of all builds are affected:

see e.g.:
http://intercity-vpn.de/bootstrap/bootstrap.log-multilog-9-1618369363.txt

PR #106, opened 2021-04-17
Untar/automake intermittent errors/GCC dependencies

stikonas:
Somewhat nicer way to bootstrap `tar`.

Instead of pre-unpacking tar sources, I added a single .c file tar unpacker.
It is then used to unpack `gzip-1.2.4.tar` and later `tar-1.12.tar.gz` (hence I swapped order of gzip and tar in the build process).

Then I also took an opportunity to convert tcc-0.9.27 into proper source package instead of pre-unpacked source. Since tcc is only distributed in bzip2 format, I delayed build of tcc-0.9.27 till quite a bit later.
Similarly, I got rid of mes libc git submodule and moved mes libc rebuild just before musl (Note that it's a newer version of mes than what we had). Mes libc script is still a bit messy (list of files to build) but slightly cleaned up compared to kaem script.

automake commits should fix some intermittent issues.
Fixes: #93
Fixes: #100

Fixes: #101 

Then there are some extra dependencies for (newer) GCC: MPFR and MPC.

PR #107, opened 2021-04-17
Fix intermittent error when bootstrapping automake-1.13.4

stikonas:
Our perl does not seem to recognize umask keyword

Fixes: #93

(Patch is just copied from automake-1.15 directory).

Actually we should also rebuild generated file `t/testsuite-part.am`


stikonas:
Closing in favour of #106

PR #111, opened 2021-04-21
Do not require chroot to be in PATH of non-root user.

stikonas:
@pder had some problems with chroot not in PATH (it was in `/sbin`).


fosslinux:
Don't really want to rely on `which` either. Can we wrap the check around a check that `which` exists?

stikonas:
> Don't really want to rely on `which` either. Can we wrap the check around a check that `which` exists?

I did it with another call to python.

Issue #112, opened 2021-04-28
automake-1.15.1: infrequently 0.5% ABORT: doc/amhello-1.0.tar.gz: recipe failed.


bittorf:
seen with 31caabd and qemu-386 with kernel 3.18.140 and 3G RAM.

see: (320 Megabytes)
http://intercity-vpn.de/bootstrap/bootstrap.log-multilog-16-1619558854.xz

```
[...]
01h14m06s | automake-1.15.1: compiling source.
01h14m06s | make[1]: Entering directory `/after/automake-1.15.1/build/automake-1.15.1'
01h14m06s | make[1]: Leaving directory `/after/automake-1.15.1/build/automake-1.15.1'
01h14m06s | called date binary with: '' in '/after/automake-1.15.1/build/automake-1.15.1'
01h14m06s | called date binary result 'Tue Apr 27 22:41:39  2021'
01h14m06s | Updating ./doc/version.texi
01h14m06s | MAKEINFO doc/automake-history.info
01h14m06s | GEN      bin/automake
01h14m06s | GEN      bin/aclocal
01h14m06s | GEN      t/ax/shell-no-trail-bslash
01h14m06s | GEN      t/ax/cc-no-c-o
01h14m07s | GEN      runtest
01h14m07s | GEN      doc/aclocal.1
01h14m07s | GEN      doc/automake.1
01h14m07s | GEN      lib/Automake/Config.pm
01h14m07s | GEN      doc/aclocal-1.15.1
01h14m07s | GEN      doc/automake-1.15.1
01h14m08s | GEN      doc/amhello-1.0.tar.gz
01h14m13s | doc/amhello-1.0.tar.gz: recipe failed.
01h14m13s | See file '/after/automake-1.15.1/build/automake-1.15.1/doc/amhello/amhello-output.tmp' for details
01h14m13s | make: *** [doc/amhello-1.0.tar.gz] Error 1
01h14m13s | # end of call src_compile:2
```

the content of the resulting file `amhello-output.tmp` is:
```
autoreconf: Entering directory `.'
autoreconf: configure.ac: not using Gettext
autoreconf: running: aclocal-1.15 --force 
autoreconf: configure.ac: tracing
autoreconf: configure.ac: not using Libtool
autoreconf: running: autoconf --force
autoreconf: running: autoheader --force
autoreconf: running: automake-1.15 --add-missing --copy --force-missing
configure.ac:7: installing './compile'
configure.ac:6: installing './install-sh'
[...]

Issue #103, opened 2021-04-15
gcc-4.0.4: infrequently 0.5% ABORT: run.sh: line 72: cd: libcpp: error 02


bittorf:
seen with a6630ff and qemu-386 with kernel 3.18.140 and 3G RAM.

see:
http://intercity-vpn.de/bootstrap/bootstrap.log-multilog-2-1618438037.txt

```
[...]
00h59m06s | checking for sysmp... no
00h59m06s | checking for getsysinfo... no
00h59m06s | checking for table... no
00h59m06s | checking for sysctl... no
00h59m06s | checking for __fsetlocking... yes
00h59m06s | checking whether canonicalize_file_name must be declared... yes
00h59m06s | checking for stdlib.h... (cached) yes
00h59m06s | checking for unistd.h... (cached) yes
00h59m06s | checking for getpagesize... (cached) yes
00h59m06s | checking for working mmap... yes
00h59m06s | checking for working strncmp... yes
00h59m06s | configure: creating ./config.status
00h59m06s | config.status: creating Makefile
00h59m06s | config.status: creating testsuite/Makefile
00h59m06s | config.status: creating config.h
00h59m06s | config.status: executing default commands
00h59m06s | run.sh: line 72: cd: libcpp: error 02
00h59m06s | Subprocess error 256
00h59m06s | ABORTING HARD
```

stikonas:
Hmm, maybe `mkdir` call failed before `cd`. We are just `cd`'ing into a directory that was just created.

So not sure why this failed, maybe it was just one off error that is not reproducible...

bittorf:
i have seen this during last mass-run 3 times...strange!

stikonas:
Hmm, so it's not one off. The code that fails is here:
https://github.com/fosslinux/live-bootstrap/blob/9f83e1213ea65cf6f2687c0c789b969cdf4cb93a/sysa/gcc-4.0.4/pass1.sh#L72

I'm not sure what we can do. Since we run with `set -e` that means `mkdir` returned with exit code 0. So I'm puzzled why cd fails.
Any ideas?

bittorf:
i will overload with an mkdir() function which always makes a sync, lets see if this helps

stikonas:
We haven't seen this recently, maybe fixed?

Issue #102, opened 2021-04-15
findutils-4.2.33: infrequently 0.5% ABORT: make: *** No targets specified and no makefile found.  Stop.


bittorf:
seen with a6630ff and qemu-386 with kernel 3.18.140 and 3G RAM.

see:
http://intercity-vpn.de/bootstrap/bootstrap.log-multilog-2-1618409001.txt

```
[...]
00h53m51s | - invoke gl_INIT in ./configure.in.
00h53m53s | /after/share/aclocal/ltdl.m4:27: warning: underquoted definition of AC_LIB_LTDL
00h53m53s | run info '(automake)Extending aclocal'
00h53m53s | or see http://sources.redhat.com/automake/automake.html#Extending-aclocal
00h53m53s | /after/share/aclocal/ltdl.m4:58: warning: underquoted definition of AC_LTDL_ENABLE_INSTALL
00h53m53s | /after/share/aclocal/ltdl.m4:68: warning: underquoted definition of AC_LTDL_SYS_DLOPEN_DEPLIBS
00h53m53s | /after/share/aclocal/ltdl.m4:94: warning: underquoted definition of AC_LTDL_SHLIBEXT
00h53m53s | /after/share/aclocal/ltdl.m4:114: warning: underquoted definition of AC_LTDL_SHLIBPATH
00h53m53s | /after/share/aclocal/ltdl.m4:126: warning: underquoted definition of AC_LTDL_SYSSEARCHPATH
00h53m53s | /after/share/aclocal/ltdl.m4:150: warning: underquoted definition of AC_LTDL_OBJDIR
00h53m53s | /after/share/aclocal/ltdl.m4:172: warning: underquoted definition of AC_LTDL_DLPREOPEN
00h53m53s | /after/share/aclocal/ltdl.m4:190: warning: underquoted definition of AC_LTDL_DLLIB
00h53m53s | /after/share/aclocal/ltdl.m4:221: warning: underquoted definition of AC_LTDL_SYMBOL_USCORE
00h53m53s | /after/share/aclocal/ltdl.m4:260: warning: underquoted definition of AC_LTDL_DLSYM_USCORE
00h53m53s | /after/share/aclocal/libtool.m4:153: warning: underquoted definition of _LT_AC_CHECK_DLFCN
00h53m53s | /after/share/aclocal/libtool.m4:508: warning: underquoted definition of _LT_AC_TRY_DLOPEN_SELF
[...]

PR #115, opened 2021-05-06
Add coreutils 8.32 and fix creation dates

melg8:
This mr will:
- add build of coreutils 8.32, 
     missing only:
     * regenerated man pages (It needs at least perl 5.8, which is not available for now)
- add bootstrap.cfg file for conditional builds variables
- add conditional, off by default, single step at the end of bootstrapping which rewrites dates to 1 jan 1970 of files and links, this is first step to achieve fully reproducible resulting file image. 
Now diffoscope shows only relevant differences of builds, and reprorts size reduced from megabytes to 2 digit k-bytes.

I want to thank friendly community in freenode IRC (#bootstrappable) - patiently answering my questions, giving me motivation to contribute to this great project.

Welcoming any suggestions and reviews of my mr.


sysa.py
```diff
@@ -441,3 +442,6 @@ def get_packages(self):
 
         # mpc 1.2.1
         self.get_file("https://mirrors.kernel.org/gnu/mpc/mpc-1.2.1.tar.gz")
+
+        # coreutils 8.32
+        self.get_file("https://mirrors.kernel.org/gnu/coreutils/coreutils-8.32.tar.gz")
```
stikonas: add end of file newline

sysa/coreutils-8.32/checksums
```diff
@@ -0,0 +1,106 @@
...
+f64449f39cb5c6ba32d7351bffc4cd336d09848be65f3faa27fa20694ec416da  /after/bin/uniq
+b5a1a7e5397d85ff47a8d28322d9b2baba4c43dc47ac31c09bd41e1c58ed5e9b  /after/bin/unlink
+493d5a768d7813042dad6f7885ca6f25071c30552c24ca77e084cf29374f3cb7  /after/bin/uptime
+73ae89aba837bc6f07a1fefdda73c6e49adea8bbe16a1ea6ef67cd7446eaa09c  /after/bin/users
+0be1a7899caf21e4597db723248463b69074208a1809f3a5b00be6ee3851442b  /after/bin/vdir
+77a269ef452faeefdff08b8d85cfa3f9c3512fb976be53346c935f99a270e6fe  /after/bin/wc
+2b4bd768ea563da974dc5386e264c8465d88cf3476196b597d1d22e59633c38c  /after/bin/who
+aa1c1f035c7290457d015ab330926115c40a0c675dcdc2c5d3476fced3fddd15  /after/bin/whoami
+107311f02be091d67a891a9e2dad23858131d46cd77e0eb8bfeebf1d77df477e  /after/bin/yes
+74a7e140f4be3f1cb2d0e0e77f6786931409f84e0140126298da12f44654540d  /after/libexec/coreutils/libstdbuf.so
```
stikonas, line 106: same here

sysa/coreutils-8.32/coreutils-8.32.sh
```diff
@@ -0,0 +1,36 @@
+# SPDX-FileCopyrightText: 2021 Melg Eight <mailto:public.melg8@gmail.com>
+#
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+src_prepare() {
```
stikonas, line 5: More work is needed here to rebuild all pre-generated files.
We don't used pre-generated configure, Makefile.in's etc...
gnulib's Makefile.am is probably auto-generated as well.

sysa/coreutils-8.32/checksums
bauen1, line 106: Nit: newline

stikonas (changes requested):

sysa/run2.sh
```diff
@@ -28,6 +29,10 @@ build mpfr-4.1.0
 
 build mpc-1.2.1
 
+build coreutils-8.32
```
bauen1, line 32: Would it make sense to push this up further so we have the `touch --date` command available if we want to reduce inconsistencies in a builds source file ?

sysa/helpers.sh
```diff
@@ -139,3 +140,11 @@ call() {
         default_"${1}"
     fi
 }
+
+# Set all files modified dates to be 1 Jan 1970.
+# Should be called at the end of bootstrapping process.
+# This functions need `touch` with support of --no-dereference
+# (at least coreutils 8.1).
+canonicalise_all_files_timestamp() {
+    find / -exec touch --no-dereference -t 197001010000.00 {} +
```
bauen1, line 149: I don't really like this hack, as it doesn't really solve the underlying issue (time difference induced differences) but only fixes a symptom.

sysa/run2.sh
melg8: Push further to where? as per https://github.com/coreutils/coreutils/releases - it is last released version?

sysa.py
melg8: Fixed in e86f74e5e611d9127b0610ff864134fde5933f91

sysa/coreutils-8.32/checksums
melg8: Fixed in e86f74e5e611d9127b0610ff864134fde5933f91

sysa/run2.sh
stikonas: We probably need fairly new automake for this, so we can't push it much further up.

bauen1: I meant if it would make sense to build it sooner, i don't think coreutils depends on mpc / gmp  ?

stikonas: Yeah, we can push it up a little bit. mpc/gmp will only be needed for newer GCC versions (and I think guile too). But it has to be after automake 1.15.

sysa/helpers.sh
stikonas: I guess it depends on how live-bootstrap might be consumed by downstream projects. Which I guess we don't really know at the moment.

melg8: I think we should differentiate between two problems:
1. One is "time difference induced differences" - like reported [1](https://github.com/fosslinux/live-bootstrap/issues/102), [2](https://github.com/fosslinux/live-bootstrap/issues/103), [3](https://github.com/fosslinux/live-bootstrap/issues/112)  in other words misbehaviour in builds 
2. noise of "changed" timestamps in filesystem
I think - proper solving of 1-st problem is very important. But 2-nd problem gets in the way.
That's why i've tried to fix 2-nd problem with available tools right now, so we can work on clear view of what's happening with 1st one. 
This is why i add this only as final step. So it is not interferes with build behaviour itself, but works as pure function, getting as input filesystem with files having random modification dates and unifies them to file system with single modification date. So forsing that "moving target" to stop.

From what i learned it's common in community to fix timestamps of dates by using fixed date at the end of creation process:
- in [nixos](https://github.com/NixOS/nix/blob/7c90552879da4d1df99b50c85e94201981e60123/src/libstore/local-store.cc#L457)
- in [debian](https://packages.debian.org/sid/strip-nondeterminism)
- i could find guix example if needed as well. 

I see this as a temporary solution, until much later we will have some more sofisticated filesystem/libc/kernel based sollution. But that solution will do same thing, replacing "real" modification date with "fake" one. 
Pros of that sollution will give us a chanse to have (after we fix non-reproducible behaviour from 1.) a chanse to check whole final filesystem hash sum individually and in CI. And this will add more benefits in development - detecting changes in file system after some tweaks of build scripts, which should not change output, for example. Because now project is missing such viable feedback, by checking only fixed amount of binaries, and not the whole thing.
So that's why i think it could serve us now, until we develop full solution in future.

I want to clear this out - i want this always be the last step of the build, so it will always only provide this last cleanup look and fill. Second, if you really want, we can have both worlds - i can change this step to be optional and off by default - so people / ci services - which need that step - at least can enable it.

sysa.py
```diff
@@ -437,11 +437,12 @@ def get_packages(self):
         # autoconf archive 2021.02.19
         self.get_file("https://mirrors.kernel.org/gnu/autoconf-archive/autoconf-archive-2021.02.19.tar.xz")
 
+        # coreutils 8.32
+        self.get_file(["https://mirrors.kernel.org/gnu/coreutils/coreutils-8.32.tar.gz",
+                      "https://git.savannah.gnu.org/cgit/gnulib.git/snapshot/gnulib-d279bc.tar.gz"])
```
stikonas, line 442: minor nitpick: align both URLs (so add one more white space on the second line)

sysa/coreutils-8.32/coreutils-8.32.sh
```diff
@@ -5,32 +5,44 @@
 src_prepare() {
     default_src_prepare
 
+    . ../../import-gnulib.sh
```
stikonas, line 8: This should go after removals, otherwise you'll remove newly imported files (which are now not pre-generated)

sysa/helpers.sh
fosslinux: I would prefer to hide this behind a optional flag. Because, this masks a few (but not all) underlying problems about non reproducible software.

parts.rst
```diff
@@ -2,6 +2,7 @@
 .. SPDX-FileCopyrightText: 2021 Andrius Štikonas <andrius@stikonas.eu>
 .. SPDX-FileCopyrightText: 2021 Paul Dersey <pdersey@gmail.com>
 .. SPDX-FileCopyrightText: 2021 fosslinux <fosslinux@aussies.space>
+.. SPDX-FileCopyrightText: 2021 Melg Eight <mailto:public.melg8@gmail.com>
```
fosslinux, line 5: Please don't include `mailto:`.

sysa/coreutils-8.32/coreutils-8.32.sh
```diff
@@ -0,0 +1,48 @@
...
+    # FORCE_UNSAFE_CONFIGURE disables "you should not run configure as root"
+    # error from configuration system of coreutils.
+
+    # enable-no-install-program disables installation of `expr` because it's
+    # not working (`expr` appears not executable file).
+
+    FORCE_UNSAFE_CONFIGURE=1 ./configure \
+        --prefix="${PREFIX}" \
+        --disable-nls \
+        --enable-no-install-program=expr \
```
stikonas, line 36: `expr` and `factor` are dynamically linked. I think we should build them instead of skipping.

I think `./configure CFLAGS="-static"` should help with that.
If not you can use `--without-gmp` which would also result in `expr` and `factor` being statically linked.

sysa/coreutils-8.32/coreutils-8.32.sh
```diff
@@ -0,0 +1,48 @@
...
+        --host=i386-unknown-linux-gnu \
+        --build=i386-unknown-linux-gnu
+}
+
+src_compile() {
+     make PREFIX="${PREFIX}" MAKEINFO="true"
+}
+
+src_install() {
+     make install PREFIX="${PREFIX}" MAKEINFO="true" DESTDIR="/after/my_test"
```
stikonas, line 47: Don't forget to fix `DESTDIR` before pushing. I assume this is test code which might be useful for you, but before PR is ready, this needs to be fixed.

parts.rst
```diff
@@ -623,3 +624,14 @@ mpc 3.2.1
 
 GNU MPC is a library for multiprecision complex arithmetic with exact rounding based
 on GNU MPFR.
+
```
stikonas, line 627: We should move coreutils description a bit higher up where it is being built.

melg8: Fixed in 7c0eb1cef653d2bc57d6902718402c769e89717f

sysa.py
melg8: Fixed in 1d8ac019d6efc776740565073e18e0cf05990d3e

parts.rst
melg8: Fixed in f4d57e2d3517cf101d5fdfd4ffa1dcda926acf6e

sysa/run2.sh
melg8: Fixed in 2258902e17dc294c46fca337708fe03373db01b6

sysa/coreutils-8.32/coreutils-8.32.sh
melg8: Fixed in 14a9ef562f2d0a66bfdc3e2a13c50d0ed320d10e

sysa/coreutils-8.32/coreutils-8.32.sh
```diff
@@ -0,0 +1,48 @@
+# SPDX-FileCopyrightText: 2021 Melg Eight <mailto:public.melg8@gmail.com>
+#
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+src_prepare() {
```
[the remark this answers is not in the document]
melg8: Fixed in 14a9ef562f2d0a66bfdc3e2a13c50d0ed320d10e

sysa/coreutils-8.32/coreutils-8.32.sh
```diff
@@ -0,0 +1,48 @@
...
+    default_src_prepare
```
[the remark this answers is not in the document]
melg8: Fixed in 6a2348539b85d220df25f310f9c561f38f7a634d

sysa/coreutils-8.32/checksums
```diff
@@ -0,0 +1,106 @@
...
+633f0aa4fc96937a55e3249c86a1a22c64369e145fc8c3d5fff9972006fcd66e  /after/bin/date
+751f832e9ce810c27c20b9b043e9ff914c6c51de14f3155cd7e9890ee9106e7d  /after/bin/dd
+c0c8fe93f33b1fd44c5b12bbeb381e98f6046b56a89bdb03f529389d5fa4233f  /after/bin/df
+760b9615a1af5e52f8af72843b5c55e652a953c612789a373b0d204f3cde95f5  /after/bin/dir
+23f604443e86eea4430a52d28c619a3596f5f19407d54c95879f4207a0af5867  /after/bin/dircolors
+a80b35b3a49540bf5387a56aeb63926594f81cbd9d4167ec4b9417559cb53285  /after/bin/dirname
+a24f69c07af2953a3c312e90d2a640bb4fb313e0ec36e2ecfeb85b6211656ce4  /after/bin/du
+e0494fc45a2abb1e46f4cfd3c04a0aa3a5e38a691115443a5542585dc496b213  /after/bin/echo
+df6b3856120316792ad15299c349ba82a397ff43a180d3662ed2ad7b6d74d322  /after/bin/env
+4a1b1c6646c1e209da5fde57b6dd9e51008b422216673ab811ba747984b0beaf  /after/bin/expand
```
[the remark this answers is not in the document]
melg8: Fixed in 0bd5d8a2c49c0d9a1fdc639e6b1f1db2bc7c9c60

sysa/helpers.sh
melg8: I've tried to implement such functionality in 33a372e722cf9b3d1e5eca6387563a147e8e8cef, suggestions are welcome.

sysa/coreutils-8.32/coreutils-8.32.sh
```diff
@@ -0,0 +1,109 @@
+# SPDX-FileCopyrightText: 2021 Melg Eight <public.melg8@gmail.com>
+#
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+# Original source code for this resides in gl subfolder inside coreutils,
+# but tar.gz archive have no such directory. We restore this files so that
+# gnulib-tool can correctly regenerate files and apply changes from gl folder.
+restore_gl_folder_in_coreutils() {
+  MODULES=(
```
fosslinux, line 9: indentation

sysa/coreutils-8.32/files/buffer-lcm
```diff
@@ -0,0 +1,27 @@
+# SPDX-FileCopyrightText: 2005-2020 Free Software Foundation, Inc.
```
fosslinux, line 1: I feel like all of these files are included in the gnulib checkout. I would prefer not to vendor these in live-bootstrap if possible.

sysa/coreutils-8.32/files/tempname.c.diff
```diff
@@ -0,0 +1,193 @@
+SPDX-FileCopyrightText: 2005-2020 Free Software Foundation, Inc.
```
fosslinux, line 1: Please add a header: what is this patch for and why do we need it? Please give it a descriptive name ending in `.patch`.

sysa/coreutils-8.32/files/tempname.diff
```diff
@@ -0,0 +1,17 @@
+SPDX-FileCopyrightText: 2005-2020 Free Software Foundation, Inc.
```
fosslinux, line 1: Ditto

sysa/coreutils-8.32/files/tempname.h.diff
```diff
@@ -0,0 +1,27 @@
+SPDX-FileCopyrightText: 2005-2020 Free Software Foundation, Inc.
```
fosslinux, line 1: This looks like it belongs with the first patch.

sysa/run2.sh
```diff
@@ -28,6 +33,11 @@ build mpfr-4.1.0
 
 build mpc-1.2.1
 
+if [ "$FORCE_TIMESTAMPS" = True ] ; then
+    echo 'Forcing all files timestamps to be 1970 year.'
```
fosslinux, line 37: How about 0 unix time, because that's really what's happening, not some arbitrary date in 1970

fosslinux (changes requested):
Reasonable, i especially like the bootstrap.cfg, very nice implementation :)

You shoudl also squash all commits APART from ' Add config file and force timestamp conditional option '

sysa/coreutils-8.32/files/buffer-lcm
melg8: I'll try to explain current problem and solution. All this files come from coreutils-8.32 packet and not from gnulib. This files are all can be found in gl subfolder of gnulib with same hash/tag as coreutils-8.32.
Problem: when they ship this coreutils in form of tar.gz they don't include thouse files. Instead, they ship generated results based on this files and call of gnulib-tool. As per  @stikonas [review](https://github.com/fosslinux/live-bootstrap/pull/115#discussion_r627727647) i can't just use those generated results, but should instead regenerate them.
So i missing like 15 o so files needed for regeneration (gnulib-tool explicitly warns that this files missing)
This files represents patches from coreutils to gnulib, which eventually should be incorporated inside gnulib.
Possible sollutions:
1. Download not just tar distributed archive with release but git repo. Drawbacks: it would require new implementation of loading stuff in sysa or usage of submodule and possibly more bugs if something not aligned with current coreutils release.
2. Just use later version of gnulib. Drawbacks: 1. not all of fixes are applied to gnulib repo. 2. later version of gnulib starts to rely on gperf. When i tried to build gperf - i found that it need g++ and not just gcc to be built.
3. Just copy-paste missing files (most of them just make files for h/c pairs and call it a day. Possible drawbacks - yea we redistribute them in repo.
So that's current disposition. So if it absolutely needed to be path 1 or worse path 2 i can rebase my mr to use ready to use stuff, but for now i don't have time and motivation to go along this paths and implement by myself. It took quite some time to figure out what's wrong with packages.

stikonas: Regarding 1. It's not necessary to have new implementation in sysa. We can download git archive of a release tag and everything will work more or less the same as releases. In fact we already do that for `sha-2-61555d.tar.gz`. Whether we want to do that for coreutils or not I don't know... @fosslinux what do you think? I'm a bit reluctant to vendor so many files in our `files/` directory, especially when they can be easily obtained elsewhere. In all other places `files` contain something like `config.h` file or some alternative implementation of some file or function. That's why I'm leaning to switching to git archive of release tag.

sysa/coreutils-8.32/files/tempname.h.diff
melg8: I've tried to merge this 3 files together, gnulib-tool don't like it:
```
mv: cannot stat `tempname.c.diff': No such file or directory
Subprocess error 256
ABORTING HARD
```

sysa/coreutils-8.32/files/tempname.c.diff
melg8: This is where this files [comes from](https://github.com/coreutils/coreutils/blob/v8.32/gl/lib/tempname.c.diff) what should i write there? i can add comment from git commit which added this file, or just write that this file is part of missing coreutils file hierarchy.

sysa/coreutils-8.32/files/tempname.diff
melg8: This is where this files [comes from](https://github.com/coreutils/coreutils/blob/v8.32/gl/lib/tempname.h.diff) what should i write there? i can add comment from git commit which added this file, or just write that this file is part of missing coreutils file hierarchy.

sysa/coreutils-8.32/files/tempname.c.diff
stikonas: Yeah, these are not our normal patches (that we put in `patches` dir). If it's upstream `diff` file, we should just ship it without modification.

sysa/coreutils-8.32/files/buffer-lcm
melg8: Maybe i found one. 5 will be - load this not from kernel.org but from github.

stikonas: That's basically what I was talking about with git archive. Although, maybe git.savannah.gnu.org is better than github for this.

sysa/coreutils-8.32/files/tempname.c.diff
melg8: This patch was removed as new solution doesn't require it. Fixed.

sysa/coreutils-8.32/files/tempname.diff
melg8: This patch was removed as new solution doesn't require it. Fixed.

sysa/coreutils-8.32/files/tempname.h.diff
melg8: This patch was removed as new solution doesn't require it. Fixed.

sysa/run2.sh
melg8: Fixed.

sysa/coreutils-8.32/files/buffer-lcm
melg8: Fixed by using git checkout version instead of release archive. But that required additional patching to fix newly discovered dependencies on generated through gperf files.

sysa/coreutils-8.32/coreutils-8.32.sh
melg8: That was removed because new implementation doesn't require this step. Fixed.

sysa/helpers.sh
melg8: Was implemented optional, disabled by default action, which could be enabled by flag (--force_timestamps) in rootfs.py.

sysa/helpers.sh
```diff
@@ -149,3 +150,11 @@ call() {
 default() {
     "default_${build_stage}"
 }
+
+# Set all files modified dates to be 0 unix time.
+# Should be called at the end of bootstrapping process.
+# This functions need `touch` with support of --no-dereference
```
stikonas, line 156: This function needs `touch` that supports --no-dereference

sysa/coreutils-8.32/patches/system.h.patch
```diff
@@ -0,0 +1,25 @@
+SPDX-FileCopyrightText: 2021 Melg Eight <public.melg8@gmail.com>
+
+SPDX-License-Identifier: GPL-3.0-or-later
+
+Remove dependency on propername.h.
+We do that, because inclusion of propername.h from gnulib brings dependencies on gperf
+generated files. We don't have gperf on this stage so we can't regenerate them.
```
stikonas, line 7: at this stage

sysa/coreutils-8.32/patches/printf.c.patch
```diff
@@ -0,0 +1,33 @@
+SPDX-FileCopyrightText: 2021 Melg Eight <public.melg8@gmail.com>
+
+SPDX-License-Identifier: GPL-3.0-or-later
+
+Remove dependency on unicodeio.h. This drops support for output unicode characters.
```
stikonas, line 5: for outputting

parts.rst
```diff
@@ -590,6 +591,17 @@ automake 1.15.1
...
 
+coreutils 8.32
+=============
+
+GNU Coreutils is a collection of widely used utilities such as ``cat``,
+``chmod``, ``chown``, ``cp``, ``install``, ``ln``, ``ls``, ``mkdir``,
+``mknod``, ``mv``, ``rm``, ``rmdir``, ``tee``, ``test``, ``true``, and
+many others.
+
+We rebuild latest available coreutils 8.32 which adds needed options to make build
```
stikonas, line 602: the latest

parts.rst
```diff
@@ -590,6 +591,17 @@ automake 1.15.1
...
+results reproducible. For example ``touch --no-dereference``.
```
stikonas, line 603: I would word it a bit differently. @fosslinux what do you think? I think this might mislead the readers a bit because they'll thing that binaries themselves are not reproducible. But it's only timestamp which is just metadata.

sysa.py
```diff
@@ -186,13 +188,23 @@ def after(self):
...
         self.mescc_tools_checksum()
         self.deploy_extra_files()
         self.mescc_tools_extra()
         self.mes()
         self.tcc_0_9_26()
         self.get_packages()
 
+    def create_configuration_file(self):
+        """
+        Creates bootstrap.cfg file which would contain options gathered
```
stikonas, line 201: I don't like console here, you can start bootstrap without console, e.g. using some GUI. Maybe just write a shorter comment like

Creates bootstrap.cfg file which would contain options used to customize bootstrap.

sysa/coreutils-8.32/coreutils-8.32.sh
```diff
@@ -0,0 +1,51 @@
+# SPDX-FileCopyrightText: 2021 Melg Eight <public.melg8@gmail.com>
+#
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+# Removing pre-generated files from source code, we will re-generate them.
```
stikonas, line 5: # Remove pre-generated files from source code

(we are not renegerating info or gmo files as we don't have texinfo or gettext yet)

sysa/coreutils-8.32/coreutils-8.32.sh
```diff
@@ -0,0 +1,51 @@
...
+remove_generated_configs() {
```
stikonas, line 6: Rename this function. It's not clear what `configs` are. Maybe `files` instead of `configs`.

sysa/coreutils-8.32/coreutils-8.32.sh
```diff
@@ -0,0 +1,51 @@
...
+    find . -name '*.info' -delete
+    find . -name '*.gmo' -delete
+
+    rm m4/gettext.m4
+}
+
+regenerate_configs() {
```
stikonas, line 13: same here. Rename.

sysa/coreutils-8.32/coreutils-8.32.sh
```diff
@@ -0,0 +1,51 @@
...
+    build-aux/gen-lists-of-programs.sh --autoconf > m4/cu-progs.m4
+    build-aux/gen-lists-of-programs.sh --automake > src/cu-progs.mk
+    build-aux/gen-single-binary.sh src/local.mk > src/single-binary.mk
+    touch ./ChangeLog
```
stikonas, line 17:
```suggestion
    touch ChangeLog
```

sysa/coreutils-8.32/coreutils-8.32.sh
```diff
@@ -0,0 +1,51 @@
...
+
+src_prepare() {
+    default
+    regenerate_configs
+}
+
+src_configure() {
+    # FORCE_UNSAFE_CONFIGURE disables "you should not run configure as root"
+    # error from configuration system of coreutils.
+    FORCE_UNSAFE_CONFIGURE=1 ./configure CFLAGS="-static -O2" \
```
stikonas, line 38: why `-O2`? We don't specify it anywhere else and I think it's the default anyway.

sysa/coreutils-8.32/patches/Makefile.am.patch
```diff
@@ -0,0 +1,18 @@
+SPDX-FileCopyrightText: 2021 Melg Eight <public.melg8@gmail.com>
+
+SPDX-License-Identifier: GPL-3.0-or-later
+
+Remove dependency on po folder. Because we lack sufficient elements
```
stikonas, line 5: Remove dependency on po folder because we don't have gettext yet.

sysa/coreutils-8.32/patches/configure.ac.patch
```diff
@@ -0,0 +1,20 @@
+SPDX-FileCopyrightText: 2021 Melg Eight <public.melg8@gmail.com>
+
+SPDX-License-Identifier: GPL-3.0-or-later
+
+Remove dependency on gettext, because we dont have it on this stage
```
stikonas, line 5: don't have it at this

' is missing and also on->at

parts.rst
```diff
@@ -590,6 +591,18 @@ automake 1.15.1
...
+coreutils 8.32
+=============
+
+GNU Coreutils is a collection of widely used utilities such as ``cat``,
+``chmod``, ``chown``, ``cp``, ``install``, ``ln``, ``ls``, ``mkdir``,
+``mknod``, ``mv``, ``rm``, ``rmdir``, ``tee``, ``test``, ``true``, and
+many others.
+
+We rebuild the latest available coreutils 8.32 which adds needed options to make
+build results metadata reproducible. For example, timestamps are changed with
```
stikonas, line 603: I think it would read better if it says `results of build metadata`.

Also maybe first paragraph is not necessary anymore? We already mentioned that earlier in coreurils 5 text.

stikonas: Oh, also maybe rebuild->build. I would reserve "rebuild" for the cases when we rebuild the same version.

stikonas (approved):
looks good to me.

@fosslinux It would be good if you can re-review too before we merge this (since it's a fairly big MR).

fosslinux (approved):

PR #122, opened 2021-05-27
Gettext/texinfo

stikonas:
* automake 1.16.3
* patch 2.7.6 (gets rid of mes libc version, now only gzip uses meslibc)
* gettext 0.21
* texinfo 6.7


fosslinux (approved):

fosslinux:
Interesting, CI ran out of space... also need to add more storage to tmpfs... Looks like not using tmpfs will be a good option.

PR #128, opened 2021-06-29
pkg-config-0.29.2

stikonas:


sysa/pkg-config-0.29.2/pkg-config-0.29.2.sh
```diff
@@ -0,0 +1,14 @@
...
+
+src_prepare() {
+    autoreconf -fi
+}
+
+src_configure() {
+    ./configure \
+        --prefix="${PREFIX}" \
+        --build=i386-unknown-linux-gnu \
+        --with-internal-glib
```
fosslinux, line 13: Could youy add a comment here explaining the reasoning for using internal glib (I know but it's not obvious reading code).

fosslinux (changes requested):

fosslinux:
`parts.rst` please. If you could also update it for other commits (I forgot about `parts.rst` when reviewing before!)

fosslinux:
Never mind I see it is ther.

PR #132, opened 2021-08-27
Kexec, Kernel, sysc, etc, etc

fosslinux:
Ok this is an absolutely humongous PR that has taken me an extremely long time to create but is really one discrete set of changes for the most part. Let's go through each of them  (and I will include some reasoning but the commit descriptions are very extensive so have a read of those).

## General stuff

**Primary goal of this PR: Adding a self-contained kernel to live-boostrap**

*Please note: commit order does not mean anything much, as my thought process and design process for this changed throughout this PR being developed, and so below I will ignore commit order as well*

## Packages added

First up is adding kexec-tools. The obvious importance of kexec is that it allows us to exec into a new kernel without user interaction. This does mean that we require `CONFIG_KEXEC` to be enabled on the seed kernel.

Other dependencies of this include linux-headers (which is required later for other purposes too). Unfortunately linux-headers is massive and we run into RAM issues in a 4GB RAM i386 host, so to clean up, there have been a few other measures introduced; linux-headers (and the Linux kernel) are manually cleaned post-build, as we need double-storage space to make a copy of `/usr` later. Also the headers do NOT have to be the same version as the kernel they are running on, so I just went for the newest at the time of making.

Another thing we need to build is kbd. Linux kernel for obvious reasons requires keymaps. However it generates these using `loadkeys` from `kbd` so we need to build that first. Newer versions = variously extremities of building problems (autoconf/gcc/etc).

Another thing we need is a gcc-built make. The make we have (tcc/mes libc) is quite buggy and is insufficient to compile the Linux kernel. Hence, we build the newest possible (autoconf reasons) make using gcc/musl, make 3.82.

The last thing we require before the linux kernel is util-linux. Technically we could build this in sysb, but it is much easier to 'separate' sysb for its own task which will be discussed shortly and so build util-linux here. The other benefit is making chroot and qemu more similar as chroot skips sysb. util-linux contains the `mount` command which is required to access the hard drive for sysc.

Before I discuss sysb and sysc, let's talk about the Linux kernel. This was the hardest part of the entire thing. Eventually I settled on Linux kernel version 4.9.10. There are a large number of binary blobs in the vanilla kernel which are unacceptable, so we 'use' Linux-Libre kernel. However, I found that the generation process for the Linux-Libre process is actually scripted. As we want to get as raw to the source as possible, we self-run the script. However, *there is an extreme disadvantage to doing this - It takes a very long time, roughly 3-4 hrs, as we must use sed as our gawk version is not new enough*. **I highly recommend for future development people should disable this check while developing.**

The kernel config is heavily modified from Void Linux. Menuconfig was intentially unused to make the config readable. Further commenting can and should be done on the config in the future to explain various choices/remove unneded things. This is IMO out of the scope of this PR. We also do not use modules as that is a significant space-eater that we do not have space for in the 4GB of RAM as will be discussed in a moment because of sysb.

The rest of the packages are essentially unchanged, apart from the more general new things changed (not package related).

## `live-bootstrap` design changes

These changes brought to the surface a number of varying severity flaws in the general design of live-bootstrap, that were previously unforseen. This took up a good majority of my time in developing this PR. (FWIW, I do not think there is any blame to apportion here, these were very hard to figure out).

While live-boostrap from the start *theoretically* supported sysb/c/d/e, there were never used and practically were very difficult to add in the current framework, due to the lack of globalisations. This lead to the first design change; the addition of sysglobal. Sysglobal contains scripts that would be needed in all systems, and a Python class inherited by all of the Sys* classes in `sys*.py`/`rootfs.py`.

This allowed for the second design change, the addition of `sysb` and `sysc`.

`sysc` contains everything after the Linux kernel. This is because we must kexec into the new kernel, which requires a re-exec of main processes and hence a new system. There is one great change that we do here though, which adds a hardware requirement to live-bootstrap; we now require a **hard drive**. sysc does not run off an initramfs. One problem that has been occuring throughout recent times in live-bootstrap is running into the 4GB RAM limit of i386, because of building larger and larger packages and the rootfs filling up. Hence, sysc now uses a hard drive. Other than that, sysc is just a continuation of sysa after the Linux kernel and is rather straightforward. sysc gets its scripts from `rootfs.py`, NOT sysa.

**WARNING!!! DO NOT RUN ON BARE METAL WITH A DISK!!! Currently sda1 is HARDCODED as the disk to use! This could overwrite a current system on disk!!**

`sysb` is a transitional system for QEMU/bare metal **only**. It is *completely skipped* in the chroot stage. The reason for this existing is because we cannot rely on the seed kernel having hard drive support, so we need to add the current binaries to sysc from our new kernel where we know we have hard drive support. This is `sysb`s only function and sysb is also run from an initramfs.

Unfortunately this brings with it the problem that we need to create the initramfs for sysb in sysa, which involves having 3 copies! of `/usr` at once (ondisk, in initramfs source, in initramfs)! This uses up a lot of the 4GB of RAM so the measures talked about earlier are used.

The last change made in this PR which is more of a nice-to-have is the movement to `/usr` from `/image`, and (in sysc only) `/after` to `/usr/src`. This makes copying files to a more normal (read FHS) Linux file system structure. We also use what is known as usrmerge which is where `/bin` is a symlink to `/usr/bin`.

Also i disabled cirrus run check because it was annoying me.

## Future work

The next things I plan to do are rather architecture-related that I have found while making this PR;

1. Add interactive configure, so that disks other than `sda` can be used. Probably along with this add a `./rootfs.py metal` target so bare metal is possible without doing random extractions from QEMU.
2. Maybe? make sysc scripts in sysa so you don't have to pre-prepare the disk.
3. Add parallelism, particularly in the Linux-libre scripts to make those faster.
4. Add `./rootfs.py snapshot` for development, so that the bootstrap starts from a snapshot tarball which is at a particular stage to allow for significantly faster development.
5. Make checksums be able to be put in a mode where they are autogened.

--- 

If you got this far thank you for reading :)


stikonas (approved):

stikonas:
Also please fix pylint and shell_lint

fosslinux:
All fixed. I will merge tomorrow if no further problems.

PR #154, opened 2022-05-04
Move from mes-m2 to mes-0.24.

stikonas:
This replaces mes-m2 with mes-0.24 release that integrated M2-Planet port.

It also includes newer libc with our fixes, so we no longer need to update meslibc later in the pipeline (and corresponding tcc rebuild is also removed).


fosslinux (approved):
just checking checksums from my pr

PR #161, opened 2022-05-14
Don't export variables from bootstrap.cfg

doraskayo:
From the main commit message:
<pre>
These variables should only affect live-bootstrap's scripts, yet they
currently "pollute" the build environment of most packages during the
bootstrap unnecessarily.
    
This change also makes bootstrap.cfg keep the same format between the
different bootstrap stages, which simplifies the input to each step.
</pre>

### DIFF
--- a/sysa/run.sh
+++ b/sysa/run.sh
@@ -72,7 +72,7 @@ if [ -z "${CHROOT}" ]; then
     ask_chroot
     echo
 
-    echo "export CHROOT=${CHROOT}" >> "${SOURCES}/bootstrap.cfg"
+    echo "CHROOT=${CHROOT}" >> "${SOURCES}/bootstrap.cfg"
 fi
 
 ask_timestamps() {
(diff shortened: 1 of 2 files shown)


doraskayo:
I'm splitting a few changes I have in a local branch into separate PRs.

They all build together just fine, but I want to ensure that they also build fine separately. So I'll keep this PR in draft form until I complete a full bootstrap alone with this change.

Note that I'm only planning to run this through a QEMU-based bootstrap, so if someone thinks this needs an additional test (bare-metal or otherwise), I'd appreicate some help in testing.

doraskayo:
> They all build together just fine, but I want to ensure that they also build fine separately. So I'll keep this PR in draft form until I complete a full bootstrap with this change alone.

It has built successfully without affecting any hash.

fosslinux:
The loop removed adds any interactively changed options to bootstrap.cfg. This is a supported case, so either the additions need to be moved into the respective `ask_` functions (probably better code decision) or the loop readded

doraskayo:
> The loop removed adds any interactively changed options to bootstrap.cfg. This is a supported case, so either the additions need to be moved into the respective `ask_` functions (probably better code decision) or the loop readded

Thanks for the review. It makes sense, I missed the interactive part.

Instead of the loop, I'll update `bootstrap.cfg` wherever a configuration is set interactively. I probably wouldn't have missed this use case if the code was written this way in the first place. Just note that `UPDATE_CHECKSUMS` doesn't seem to provide an interactive prompt, so it will never be updated.

I'll also try to run an interactive bootstrap with QEMU to confirm it works as expected.

doraskayo:
> I'll also try to run an interactive bootstrap with QEMU to confirm it works as expected.

I confirmed that both interactive and non-interactive bootstraps work after my last changes.

stikonas:
> Just note that `UPDATE_CHECKSUMS` doesn't seem to provide an interactive prompt, so it will never be updated.

`UPDATE_CHECKSUMS` is development aid, so I deliberately didn't add it there.

stikonas (approved):

PR #162, opened 2022-05-14
Fix libtool archive reproducibility

doraskayo:
From the commit message:
<pre>
This makes the order of objects in archives created by libtool
consistent.

It is known to affect cases where the *_LIBADD automake variable is
used to add extra objects from a separate archive.

Fixing this allows us to remove a few workarounds.
</pre>


stikonas (approved):

fosslinux (approved):

fosslinux:
had no idea what was causing this weird behaviour, thank you for the fix

PR #163, opened 2022-05-14
Additional preparations for rootless bootstrap

doraskayo:
This MR is similar to https://github.com/fosslinux/live-bootstrap/pull/144, but handles a few issues that came up since.

### DIFF
--- a/sysc/run.sh
+++ b/sysc/run.sh
@@ -18,11 +18,11 @@ create_fhs() {
     for d in bin lib sbin; do
         ln -s "usr/${d}" "/${d}"
     done
-    mkdir -p /etc /proc /run /sys /tmp /var
-    mount -t proc proc /proc
-    mount -t sysfs sysfs /sys
+    mkdir -p /etc /run /var
+    test -d /proc || (mkdir /proc && mount -t proc proc /proc)
+    test -d /sys || (mkdir /sys && mount -t sysfs sysfs /sys)
     # Make /tmp a ramdisk (speeds up configure etc significantly)
-    mount -t tmpfs tmpfs /tmp
+    test -d /tmp || (mkdir /tmp && mount -t tmpfs tmpfs /tmp)
 }
 
 populate_device_nodes ""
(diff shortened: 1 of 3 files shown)


doraskayo:
I'm splitting a few changes I have in a local branch into separate PRs.

They all build together just fine, but I want to ensure that they also build fine separately. So I'll keep this PR in draft form until I complete a full bootstrap with this change alone.

sysc/run.sh
```diff
@@ -24,11 +24,11 @@ create_fhs() {
     for d in bin lib sbin; do
         ln -s "usr/${d}" "/${d}"
     done
-    mkdir /etc /proc /run /sys /tmp /var
-    mount -t proc proc /proc
-    mount -t sysfs sysfs /sys
+    mkdir /etc /run /var
+    test -d /proc || (mkdir /proc && mount -t proc proc /proc)
```
fosslinux, line 28: Alternatively, `mkdir -p` for slightly nicer code

doraskayo: I think `mkdir -p` wouldn't get the expected result.

For some background, I plan to mount these filesystems before the bootstrap begins as part of setting up the bubblewrap-based sandbox setup.

So in this function I want to checkk if they are already mounted, or need to be mounted. This is similar to the logic in [`populate_device_nodes`](https://github.com/fosslinux/live-bootstrap/blob/fba7ecf96a11c74063e3d1208f6b85147d9bf32a/sysa/helpers.sh#L369).

However, if I create the directories unconditionally using `mkdir -p`, I would always thing that they exist and would never execute `mknod`.

Please let me know if it makes sense, or if I missed something in your suggestion.

doraskayo:
> They all build together just fine, but I want to ensure that they also build fine separately. So I'll keep this PR in draft form until I complete a full bootstrap with this change alone.

The build succeeded without changing any hashes.

doraskayo:
Fixed conflicts with `master`.

fosslinux (approved):

PR #173, opened 2022-05-21
Fix file permissions for run2.kaem

doraskayo:
This is a small follow-up fix for https://github.com/fosslinux/live-bootstrap/pull/171.

It seems that `chmod +x` doesn't work as expected at this stage of the bootstrap, and instead removes all file permissions. This prevents rootless bootstrap from working, but doesn't seem to affect other bootstrap modes.


stikonas (commented):
Maybe let's remove chmod completely here?

PR #174, opened 2022-05-22
Allow preseeding with binary packages.

stikonas:
There is also a new rootfs.py option to specify previous binary repo.

### DIFF
--- a/sysa.py
+++ b/sysa.py
@@ -34,7 +34,7 @@ class SysA(SysGeneral):
         self.sysb_dir = sysb_dir
         self.sysc_tmp = sysc_tmp
 
-    def prepare(self, mount_tmpfs, copy_sysc, create_initramfs):
+    def prepare(self, mount_tmpfs, copy_sysc, create_initramfs, repo_path):
         """
         Prepare directory structure for System A.
         We create an empty tmp directory, unpack stage0-posix.
@@ -53,6 +53,10 @@ class SysA(SysGeneral):
 
         if copy_sysc:
             self.sysc()
+
+        if repo_path:
+            repo_dir = os.path.join(self.tmp_dir, 'usr', 'src', 'repo-preseeded')
+            shutil.copytree(repo_path, repo_dir)
 
         if create_initramfs:
             self.make_initramfs()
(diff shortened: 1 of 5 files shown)


stikonas:
This is now ready for review.

sysa.py
```diff
@@ -37,6 +37,10 @@ def __init__(self, arch, preserve_tmp, tmpdir, chroot, sysb_dir, sysc_tmp):
 
         self.prepare()
 
+        if repo_path:
+            repo_dir = os.path.join(self.tmp_dir, 'usr', 'src', 'repo-preseeded')
+            shutil.copytree(repo_path, repo_dir)
```
doraskayo, line 42: The `copytree` will need to be moved to the `prepare()` method after my PR (https://github.com/fosslinux/live-bootstrap/pull/175).

The intent being that the sysx constructors are only used to set options that apply to all bootstrap modes, but the actual prepartion operations such as creating directories, mounting them and copying files are done in `prepare()` and initiated by each bootstrap mode in a way that fits its expecations and capabilities.

As such. the `self.tmp_dir` directory will not be created in the context of the constructor, and therefore the `shutil.copytree` line cannot happen in the constructor either.

sysa/helpers.sh
```diff
@@ -86,6 +101,28 @@ build() {
...
+        cd "${SRCDIR}/repo-preseeded"
+        if src_checksum "${pkg}" $((revision)); then
+            echo "${pkg}: installing prebuilt package."
+            if [[ "${pkg}" == bash-* ]]; then
+                # tar does not like overwriting running bash
+                # shellcheck disable=SC2153
+                rm -f "${PREFIX}/bin/bash" "${PREFIX}/bin/sh"
+            fi
+            mv "${pkg}_${revision}"* ../repo
+	    # shellcheck disable=SC2144
```
fosslinux, line 115: Indentation

sysa/helpers.sh
```diff
@@ -86,6 +101,28 @@ build() {
...
+            if [ -f *-repodata ]; then
+                mv -- *-repodata ../repo
+            fi
+            cd "${SRCDIR}/repo"
+            src_apply "${pkg}" $((revision))
+            cd "${SOURCES}"
+            return
+        fi
+    fi
```
fosslinux, line 124: Split this out into another function in keeping with the style of build()?

fosslinux (approved):

PR #175, opened 2022-05-23
Add a rootless bootstrap mode using bubblewrap

doraskayo:
This PR adds the ability to bootstrap using the `bwrap` utility, which can operate in rootless mode when user namespaces are supported.

Most of the commits in this PR rework the Python scripts to allow better integration of additional bootstrap modes, and some fix existing issues. The most significant change is that the `prepare()` methods of sysa and sysc are now called directly from `rootfs.py`, and provided with the configuration that each bootstrap mode expects.

### DIFF
--- a/rootfs.py
+++ b/rootfs.py
@@ -109,10 +109,10 @@ def main():
             pass
 
     system_c = SysC(arch=args.arch, preserve_tmp=args.preserve,
-            tmpdir=args.tmpdir, chroot=args.chroot)
+                    tmpdir=args.tmpdir)
     system_b = SysB(arch=args.arch, preserve_tmp=args.preserve)
     system_a = SysA(arch=args.arch, preserve_tmp=args.preserve,
-                    tmpdir=args.tmpdir, chroot=args.chroot,
+                    tmpdir=args.tmpdir,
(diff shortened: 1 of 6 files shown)


rootfs.py
```diff
@@ -139,6 +143,45 @@ def bootstrap(args, system_a, system_b, system_c):
         init = os.path.join(os.sep, 'bootstrap-seeds', 'POSIX', arch, 'kaem-optional-seed')
         run('sudo', 'env', '-i', 'PATH=/bin', chroot_binary, system_a.tmp_dir, init)
 
+    elif args.bwrap:
+        find_bwrap = """
+import shutil
+print(shutil.which('bwrap'))
+"""
```
stikonas, line 150: Do we need this hack for `bwrap`? Can we not rely on `PATH`?

for `chroot` we couldn't rely on PATH because `chroot` might be in sbin which is not in PATH.

Also as discussed, sort out removing of tmpdir.

Rest of the stuff looks good to me.

fosslinux:
Ohter than what stikonas pointed out, lgtm

doraskayo:
Thanks for the review.

The `which` hack was removed, and `tmp` directories are now removed recursively. I also added myself under `SPDX-FileCopyrightText`.

stikonas (approved):

PR #178, opened 2022-05-30
Network

fosslinux:
Add the ability to download files from the network.

Still todo (but not in this already large PR)
- be able to custom specify mirror server to download from (as stikonas requested)
- static ips. only dhcp ips are currently supported through dhcp.

All previous behaviour has been retained as an option.

Not all testing is 100% complete.

All before curl rebuild with openssl must be http only. This lead to a few tarball changes.


sysa/curl-7.83.0/patches/help.patch
```diff
@@ -0,0 +1,55 @@
+diff --color -ru src/Makefile.inc src/Makefile.inc
```
stikonas, line 8: SPDX header and a comment with explanation why this is needed

sysa/curl-7.83.0/files/sinus.pl
```diff
@@ -0,0 +1,4 @@
+my $pi = 3.1415;
```
stikonas, line 5: SPDX header

sysc/curl-7.83.0/files/sinus.pl
```diff
@@ -0,0 +1,4 @@
+my $pi = 3.1415;
```
stikonas, line 5: SPDX header

sysc/curl-7.83.0/patches/help.patch
```diff
@@ -0,0 +1,55 @@
+diff --color -ru src/Makefile.inc src/Makefile.inc
```
stikonas, line 8: same as in sysa, spdx header and comment

stikonas (approved):
Looks good. There might be a few improvements possible but maybe better to split them into separate PRs.
* TODOs that are already mentioned here (mirror, etc).
* If possible move openssl and curl rebuild as early as possible in sysc (unless openssl depends on something late).

sysc.py
```diff
@@ -133,22 +137,26 @@ def get_packages(self):
...
 
         # libarchive-3.5.2
         self.get_file("https://libarchive.org/downloads/libarchive-3.5.2.tar.xz")
 
         # openssl-1.1.1l
-        self.get_file("https://www.openssl.org/source/openssl-1.1.1l.tar.gz")
+        self.get_file("https://artfiles.org/openssl.org/source/old/1.1.1/openssl-1.1.1l.tar.gz")
+
+        # curl 7.83.0
+        self.get_file("https://master.dl.sourceforge.net/project/curl.mirror/curl-7_83_0/curl-7.83.0.tar.xz?viasf=1")
```
stikonas, line 153: Can we output this into a nicer location?

stikonas (changes requested):

stikonas (changes requested):
When running `./rootfs.py -bw --external-sources` I get some errors like

```
cp: cannot open `/sysc/sys/kernel/slab/:A-0000128/trace' for reading: Permission denied
cp: cannot open `/sysc/sys/kernel/slab/:A-0000128/reclaim_account' for reading: Permission denied
cp: cannot open `/sysc/sys/kernel/slab/:A-0000128/order' for reading: Permission denied
cp: cannot open `/sysc/sys/kernel/slab/:A-0000128/object_size' for reading: Permission denied
cp: cannot open `/sysc/sys/kernel/slab/:A-0000128/shrink' for reading: Permission denied
cp: cannot open `/sysc/sys/kernel/slab/:A-0000128/hwcache_align' for reading: Permission denied
cp: cannot open `/sysc/sys/kernel/slab/:A-0000128/objs_per_slab' for reading: Permission denied
cp: cannot open `/sysc/sys/kernel/slab/:A-0000128/partial' for reading: Permission denied
cp: cannot open `/sysc/sys/kernel/slab/:A-0000128/slabs_cpu_partial' for reading: Permission denied
cp: cannot open `/sysc/sys/kernel/slab/:A-0000128/ctor' for reading: Permission denied
cp: cannot open `/sysc/sys/kernel/slab/:A-0000128/slab_size' for reading: Permission denied
cp: cannot open `/sysc/sys/kernel/slab/:A-0000128/cache_dma' for reading: Permission denied
cp: cannot create special file `/sysc_image//usr/src/sys/kernel/security/apparmor/.null': Operation not permitted
```
(more is printed but I truncated it)

sysc.py
fosslinux: oh I will just outright remove the `?viasf=1` now as `-L` added to curl which follows redirects (which is why it was there in first place).

rootfs.py
```diff
@@ -162,22 +167,21 @@ def bootstrap(args, system_a, system_b, system_c):
                      '--uid', '0',
                      '--gid', '0',
                      '--cap-add', 'CAP_SYS_CHROOT', # Required for chroot from sysa to sysc
-                     '--clearenv',
```
stikonas, line 165: Why are we removing `clearenv`? I think it can be readded.

sysc.py
```diff
@@ -133,22 +137,26 @@ def get_packages(self):
...
+        # curl 7.83.0
+        self.get_file("https://master.dl.sourceforge.net/project/curl.mirror/curl-7_83_0/curl-7.83.0.tar.xz")
+
+        # ca-certificates-3.78
+        self.get_file("https://ftp.mozilla.org/pub/security/nss/releases/NSS_3_78_RTM/src/nss-3.78.tar.gz")
 
         # xbps 0.59.1
-        self.get_file("https://github.com/void-linux/xbps/archive/refs/tags/0.59.1.tar.gz",
-                       output="xbps-0.59.1.tar.gz")
+        self.get_file("https://github.com/void-linux/xbps/archive/refs/tags/0.59.1.tar.gz")
```
stikonas, line 159: Hmm, it's a bit ugly to download tarball into `0.59.1.tar.gz`. `xbps-0.59.1.tar.gz` was nicer.

stikonas (changes requested):
just a couple of minor points.

sysc.py
fosslinux: I agree, but there's not really any mechanism to change the output filename when downloaded in sysc right now (nor can I figure out an elegant solution). We can look into that a bit later

stikonas: Ok, let's merge it like this then. Maybe add a `TODO` comment and merge?

PR #189, opened 2022-09-02
Create a symlink for GCC

doraskayo:
From the commit message:
```
This is required when the "CC" environment variable isn't set
explicitly, which is to be expected when using sysc_image in
downstream projects.
```


fosslinux (approved):
Reasonable change :+1:

PR #190, opened 2022-09-03
Fix broken link to stage0-posix in parts.rst

TonalidadeHidrica:
There is an invalid link in `parts.rst`, so I guessed the intention and fixed it.


fosslinux (approved):
This is a relic from when stage0-posix was all in on repo.

PR #191, opened 2022-09-04
Add binutils 2.38 and libtool 2.4.7

doraskayo:
## Overview

This PR adds `binutils 2.38` to the bootstrap process. Introducing `libtool 2.4.7` as well helped avoid configuration issues, and building after `gcc 4.7.4` allowed a minimal amount of patching and workarounds during compilation.

`libtool` is built as early as its minimal requirements are met. It did require a small workaround in `gcc 4.7.4` caused by this package relying upon an older `automake` version relatively late in the bootstrap. The workaround did not affect the actual resulting package.

I chose `binutils 2.38` because I already build it successfully on the `freedesktop-sdk` side, and wanted to be on the safe side when attempting to bootstrap it in `live-bootstrap`. I assume it should be possible to update to `2.39` without much trouble in the future.

## Testing

I used both the `qemu` and `bwrap` bootstrap modes in my tests. I bootstrapped each commit individually, and verified that hashes remain stable through multiple bootstrap exectutions.

cc: @stikonas, @fosslinux


sysc/binutils-2.38/binutils-2.38.sh
```diff
@@ -0,0 +1,86 @@
+# SPDX-FileCopyrightText: 2022 Dor Askayo <dor.askayo@gmail.com>
+# SPDX-FileCopyrightText: 2021 Andrius Štikonas <andrius@stikonas.eu>
+# SPDX-FileCopyrightText: 2021 Paul Dersey <pdersey@gmail.com>
+#
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+urls="https://mirrors.kernel.org/gnu/binutils/binutils-2.38.tar.bz2"
```
fosslinux, line 7: Here we can use xz, afaict

sysc/libtool-2.4.7/libtool-2.4.7.sh
```diff
@@ -0,0 +1,47 @@
...
+
+src_prepare() {
+    default
+
+    rm -f build-aux/ltmain.sh
+    rm -f doc/*.info
+
+    mv ../gnulib-a521820 gnulib
+
+    AUTOMAKE=automake-1.11 ACLOCAL=aclocal-1.11 AUTOM4TE=autom4te-2.64 AUTOCONF=autoconf-2.64 AUTOHEADER=autoheader-2.64 AUTORECONF=autoreconf-2.64 ./bootstrap --force
```
fosslinux, line 18: This bootstrap script appears to be autogened itself.

fosslinux (commented):
This seems to be good generally. I would like to do a pass over the tarball myself for any pregen things missed first however.

sysc/libtool-2.4.7/libtool-2.4.7.sh
doraskayo: Apparently we have a similar issue here:
https://github.com/fosslinux/live-bootstrap/blob/master/sysa/libtool-2.2.4/libtool-2.2.4.sh#L13

doraskayo: @fosslinux, I managed to bootstrap the `bootstrap` script, but so far it only works with `--external-sources`.

`libtool` seems to be using a git snapshot of the `bootstrap` project, whose upstream is https://github.com/gnulib-modules/bootstrap. Unfortunately, as far as I can tell, `github.com` doesn't allow HTTP downloads for its tarballs; it redirects to HTTPS. Our curl at this point in the bootstrap can't handle HTTPS.

This is the relevant tarball: https://github.com/gnulib-modules/bootstrap/archive/3d6394c3617a6a79e460ac513fa3281f8ae72cd0.tar.gz

What do you suggest we do? Should I build `libtool 2.4.7` only after `curl` is rebuilt with HTTPS support? It would be a bit unfortunate.

stikonas: I don't think we have an issue with libtool 2.2.4. At least that script from 2.2.4 seems handwritten to me.

stikonas: hmm, can we build `curl` with HTTPS support earlier? Not sure if other libraries are easier to build then `openssl`, e.g. `mbedtls` or others.

doraskayo: I ended up extracting the minimal set of operations performed by the `bootstrap` script to get an identical final package. While it ended up slightly annoying due to a specific hook that `libtool` implements during the bootstrap script's execution, I created a small script (`bootstrap-helper.sh`) to hopefully document it better and make future updates easier.

The `bootstrap` script itself is now unused and deleted at the prepartion stage, so I'm marking this thread as resolved.

sysc/binutils-2.38/binutils-2.38.sh
doraskayo: Indeed.

stikonas (approved):
Looks good to me but let's wait for @fosslinux to review tarball for pregenerated files.

sysc/binutils-2.38/binutils-2.38.sh
```diff
@@ -0,0 +1,86 @@
...
+    rm gas/itbl-parse.h gold/yyscript.c
+    rm gold/yyscript.h intl/plural.c
+    rm ld/deffilep.c ld/deffilep.h
+    rm ld/ldgram.c ld/ldgram.h
+
+    # Rebuild flex generated files
+    touch -- */*.l */*/*.l
+    rm binutils/arlex.c binutils/deflex.c binutils/syslex.c
+    rm gas/config/bfin-lex.c gas/config/loongarch-lex.c gas/itbl-lex.c
+    rm ld/ldlex.c
```
fosslinux, line 47: Also `bfd/mep-relocs.pl` should be run, it won't actually change anything but verifies continuity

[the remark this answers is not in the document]
stikonas: Hmm, this also applies to binutils-2.14

sysc/binutils-2.38/binutils-2.38.sh
```diff
@@ -0,0 +1,86 @@
...
+            --build=i386-unknown-linux-gnu \
+            --host=i386-unknown-linux-gnu \
+            --target=i386-unknown-linux-gnu \
+            --program-prefix="" \
+            --with-sysroot="${PREFIX}" \
+            --prefix="${PREFIX}" \
+            --libdir="${PREFIX}/lib/musl" \
+            --srcdir=.
+        cd ..
+    done
```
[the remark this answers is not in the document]
doraskayo: Done.

sysc/binutils-2.38/binutils-2.38.sh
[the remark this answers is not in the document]
doraskayo: I removed the pregenerated files from `opcodes`, but it seems that they aren't regenerated automatically during the build unless in maintainer mode. So I added a small patch to ensure the files we actually need are generated. This was enough to get everything working.

doraskayo: Done.

doraskayo:
@fosslinux, as expected, there wasn't any change in package hashes after regenerating the various pregenerated files you mentioned.

sysc/binutils-2.38/binutils-2.38.sh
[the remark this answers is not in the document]
doraskayo: I also created a PR for `binutils 2.14`: https://github.com/fosslinux/live-bootstrap/pull/195

stikonas (approved):

stikonas:
We missed at least one pregenerated file...

`libtool/lib/uniwidth/width0.h` `libtool/lib/uniwidth/width2.h`

PR #194, opened 2022-09-11
musl dynamic linking support

doraskayo:
## Overview

This PR enables a native musl toolchain by patching `gcc 4.7.4`. Combined with `binutils 2.38` (https://github.com/fosslinux/live-bootstrap/pull/191), it allows dynamic linking in musl to work without a need for musl-specific wrapper scripts. `musl 1.2.3` is then rebuilt to actually enable dynamic linking and create `libc.so`, which also acts as musl's dynamic linker.

The GCC patches that achieve this were backported by me from GCC upstream without any additional changes.

## Result

``` cpp
bash-5.1# cat test.cc 
#include <iostream>

int main(int argc, char **argv)
{
	std::cout << "Hello World" << std::endl;
	return 0;
}
```
``` bash
bash-5.1# g++ test.cc -o test-musl-dynamic
bash-5.1# ./test-musl-dynamic 
Hello World
bash-5.1# readelf -d test-musl-dynamic | grep NEEDED
 0x00000001 (NEEDED)                     Shared library: [libc.so]
bash-5.1# readelf -a test-musl-dynamic | grep interpreter                                           
      [Requesting program interpreter: /lib/ld-musl-i386.so.1]
```

In addition to the above, this PR got me further towards a fully working Python 3 in the `freedesktop-sdk` side (Python is necessary for `glibc` and newer `gcc`). This change allows it to load and execute built-in modules successfully at runtime (requires support for `dlopen`, which in turn requires support for dynamic linking). Only a few built-in modules are now missing to get it fully working.

## Non-goals

* Some changes in the backported patches were not included as described in the commit message, mostly because they are irrelavant to `live-bootstrap` at this time, but also because they had conflicts. Building a newer GCC version would likely be a better approach to get them "for free".
* I haven't investigated whether a native toolchain can be achieved earlier in the bootstrap using older `binutils` or `gcc` versions. It may be possible, but exploring this was outside my intended scope.
* Removing the `--disable-shared` option from various packages that follow wasn't done as part of this PR. This can be explored in a future change.

## Testing

This was tested using the `bwrap` and `qemu` bootstrap modes. Each commit was tested individually and includes its package hash changes.

## Dependencies

Depends on https://github.com/fosslinux/live-bootstrap/pull/191 and therefore includes it.

cc: @stikonas, @fosslinux


sysc/musl-1.2.3/musl-1.2.3.sh
```diff
@@ -0,0 +1,24 @@
...
+        --host=i386-unknown-linux-musl \
+        --prefix="${PREFIX}" \
+        --libdir="${PREFIX}/lib/musl" \
+        --includedir="${PREFIX}/include/"
+
+    # configure script creates this file
+    if test -f /dev/null; then
+        rm /dev/null
+        mknod -m 666 /dev/null c 1 3
+    fi
```
stikonas, line 19: Could you please check if this workaround still necessary. I still don't understand what causes it but it could be `bash` built with meslibc.

fosslinux (approved):
lgtm

sysc/musl-1.2.3/musl-1.2.3.sh
doraskayo: I removed this logic locally, used the `qemu` bootstrap mode and `/dev/null` seemed fine at the end of the bootstrap:
```
bash-5.1# stat /dev/null
  File: /dev/null
  Size: 0         	Blocks: 0          IO Block: 4096   character special file
Device: 801h/2049d	Inode: 14          Links: 1     Device type: 1,3
Access: (0666/crw-rw-rw-)  Uid: (    0/ UNKNOWN)   Gid: (    0/ UNKNOWN)
Access: 2022-09-13 12:34:27.657000000 +0000
Modify: 2022-09-13 12:34:27.657000000 +0000
Change: 2022-09-13 12:34:27.657000000 +0000
 Birth: -
```

So I guess there's indeed no need for this workaround. I removed it.

sysc/musl-1.2.3/musl-1.2.3.sh
```diff
@@ -0,0 +1,18 @@
+# SPDX-FileCopyrightText: 2022 Dor Askayo <dor.askayo@gmail.com>
+# SPDX-FileCopyrightText: 2022 Andrius Štikonas <andrius@stikonas.eu>
+#
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+urls="https://musl.libc.org/releases/musl-1.2.3.tar.gz"
+
+src_configure() {
+    ./configure \
```
doraskayo, line 9: It seems that `musl` installing its dynamic linker binary into `/lib` causes `xbps-install` to remove the `/lib`->`usr/lib` symlink and make `/lib` a normal directory. We need better handling here.

doraskayo: I ended up moving the symlink into `/usr/lib` like other distros seem to do. I also made it dynamic while at it.

doraskayo:
The latest revision has some fixes and improvements:
1. Library search path configuration. This was missing in the previous revisions.
2. Changing the dynamic linker's symlink location to `/usr/lib` to avoid overriding the `/lib -> usr/lib` symlink. I also made it relative while at it.
3. Addition of `ldd`, which is simply a symlink to the dynamic linker.
   ``` bash
   bash-5.1# ldd ./test-musl-dynamic                  
	   /lib/ld-musl-i386.so.1 (0xf7e76000)
	   libtest.so => /usr/lib/musl/libtest.so (0xf7e6b000)
	   libc.so => /lib/ld-musl-i386.so.1 (0xf7e76000)
   ```

Hopefully this means all the issues were resolved and this PR should be acceptable for merging.

PR #197, opened 2022-09-16
Avoid exporting variables in build scripts

doraskayo:
## Overview
From the main commit message:
    
    Environment variables "pollute" the build environment of packages and
    can affect their output.
    
    This change results in the removal of some files from packages that
    were not meant to be packaged. It also removes the need for a
    workaround in automake 1.10.3 to manually remove such files.
    
    Variables are now saved in an .env file for each system and included
    in scripts that need them using the dot operation.

The following files were removed from `gettext-0.21_0.tar.bz2`, resulting in a change in its hash:
``` diff
+++ after-gettext-0.21_0.tar.bz2
│   --- before-gettext-0.21_0.tar
├── +++ after-gettext-0.21_0.tar
│ ├── file list
│ │ @@ -2054,18 +2054,7 @@
│ │  -rw-r--r--   0        0        0       19 1970-01-01 00:00:00.000000 ./usr/share/man/man3/dcgettext.3
│ │  -rw-r--r--   0        0        0       20 1970-01-01 00:00:00.000000 ./usr/share/man/man3/dcngettext.3
│ │  -rw-r--r--   0        0        0       19 1970-01-01 00:00:00.000000 ./usr/share/man/man3/dgettext.3
│ │  -rw-r--r--   0        0        0       20 1970-01-01 00:00:00.000000 ./usr/share/man/man3/dngettext.3
│ │  -rw-r--r--   0        0        0     4788 1970-01-01 00:00:00.000000 ./usr/share/man/man3/gettext.3
│ │  -rw-r--r--   0        0        0     2740 1970-01-01 00:00:00.000000 ./usr/share/man/man3/ngettext.3
│ │  -rw-r--r--   0        0        0     2172 1970-01-01 00:00:00.000000 ./usr/share/man/man3/textdomain.3
│ │ -drwxr-xr-x   0        0        0        0 1970-01-01 00:00:00.000000 ./usr/src/
│ │ -drwxr-xr-x   0        0        0        0 1970-01-01 00:00:00.000000 ./usr/src/gettext-0.21/
│ │ -drwxr-xr-x   0        0        0        0 1970-01-01 00:00:00.000000 ./usr/src/gettext-0.21/build/
│ │ -drwxr-xr-x   0        0        0        0 1970-01-01 00:00:00.000000 ./usr/src/gettext-0.21/build/gettext-0.21/
│ │ -drwxr-xr-x   0        0        0        0 1970-01-01 00:00:00.000000 ./usr/src/gettext-0.21/build/gettext-0.21/gettext-tools/
│ │ -drwxr-xr-x   0        0        0        0 1970-01-01 00:00:00.000000 ./usr/src/gettext-0.21/build/gettext-0.21/gettext-tools/src/
│ │ -drwxr-xr-x   0        0        0        0 1970-01-01 00:00:00.000000 ./usr/src/gettext-0.21/build/gettext-0.21/gettext-tools/src/textstyle/
│ │ --rw-r--r--   0        0        0     4331 1970-01-01 00:00:00.000000 ./usr/src/gettext-0.21/build/gettext-0.21/gettext-tools/src/textstyle/stdbool.h
│ │ --rw-r--r--   0        0        0     1348 1970-01-01 00:00:00.000000 ./usr/src/gettext-0.21/build/gettext-0.21/gettext-tools/src/textstyle/version.h
│ │ --rw-r--r--   0        0        0     1155 1970-01-01 00:00:00.000000 ./usr/src/gettext-0.21/build/gettext-0.21/gettext-tools/src/textstyle/woe32dll.h
│ │ --rw-r--r--   0        0        0    22817 1970-01-01 00:00:00.000000 ./usr/src/gettext-0.21/build/gettext-0.21/gettext-tools/src/textstyle.h
```

## Testing
This was tested in the `bwrap` and `qemu` bootstrap modes.


sysc/run.sh
```diff
@@ -8,6 +8,17 @@
 
 set -e
 
+if [ "$#" -ne 5 ]; then
+    echo "Usage: run.sh <prefix> <sources> <destdir> <distfiles> <srcdir>"
+    exit 1
+fi
```
doraskayo, line 14: I wasn't sure if it was a good idea to keep these checks to more easily detect issues during development. A failrue to pass all the expected variables would break things in unexpected ways.

I'd appreciate some opinions on this.

sysb/run.sh
```diff
@@ -74,8 +82,6 @@ if [ $(($(ls -l "/dev/${DISK}" | sed "s/.*, *//" | sed "s/ .*//") % 8)) -eq 0 ];
 fi
 echo "export DISK=${DISK}" >> /usr/src/bootstrap.cfg
 
-PREFIX=/usr
-SOURCES="${PREFIX}/src"
 SYSC=/sysc
 
 # Otherwise, add stuff from sysa to sysb
```
doraskayo, line 87: Testing with `qemu`, it seems there's a build failure in `sysb`.

sysc/run.sh
fosslinux: I think this makes sense. However a comment above this to this effect would be appreciated, along with the output when there is an error noting that this script shouldn't really be called directly (apart from during development).

sysb/run.sh
doraskayo: Never mind, the issue was with my DNS server.

sysc/run.sh
doraskayo: Sure. Done.

sysc/run2.sh
```diff
@@ -115,6 +115,7 @@ echo "Bootstrapping completed."
 cd "/"
 
 if [ -e after.sh ]; then
-    FILE=after.sh
+    exec env -i PATH="${PATH}" PREFIX="${PREFIX}" SOURCES="${SOURCES}" DESTDIR="${DESTDIR}" DISTFILES="${DISTFILES}" bash ${FILE}
+else
+    exec env -i PATH="${PATH}" bash
```
stikonas, line 135: Can we save those environmental variables to some file that can be sourced for development purposes.

sysa/run.sh
```diff
@@ -9,13 +9,13 @@
...
 # shellcheck disable=SC2154
-export PREFIX="${prefix}"
+PREFIX="${prefix}"
 # shellcheck disable=SC2154
-export SOURCES="${sysa}"
-export DISTFILES="${sysa}/distfiles"
-export DESTDIR=/tmp/destdir
+SOURCES="${sysa}"
+DISTFILES="${sysa}/distfiles"
+DESTDIR="/tmp/destdir"
```
stikonas, line 16: `tmp/destdir` should be fine, there are no variables here.

sysb/init
```diff
@@ -7,9 +7,10 @@
 
 set -e
 
-export PATH=/usr/bin
-export PREFIX=/usr
-export SOURCES=/usr/src
+export PATH="/usr/bin"
+
+PREFIX="/usr"
```
stikonas, line 12: `/usr`

sysc/run2.sh
doraskayo: Done. They are now saved in an `.env` (dotenv) file and included in scripts using the dot operation.

sysa/run.sh
doraskayo: Removed all unnecessary quotation marks.

sysb/init
doraskayo: Removed all unnecessary quotation marks.

doraskayo:
I decided to skip https://www.shellcheck.net/wiki/SC2129; I'm personally not a fan of this style. @fosslinux, let me know if it's fine.

stikonas (approved):

fosslinux (approved):
I agree on shellcheck. (I actually prefer `cat >> blah <<EOF` over either of those, but this is fine as-is).

PR #206, opened 2022-09-25
Autogen

stikonas:
Should be good enough to start review (it might take some time).

I still need to resolve non-reproducibility (https://github.com/schierlm/gnu-autogen-bootstrapping/issues/10), but I might just do it in `gnu-autogen-bootstrapping` and then pull in new tag.


stikonas:
I think all reproducibility issues are now resolved (including those that only occur on year change).

fosslinux (approved):
LGTM. I will fix up sources stuffs in my pr

Issue #188, opened 2022-08-05
Split stage sources to parseable configuration files


nanonyme:
Currently downloaded URL's are stored as Python unless I'm missing something. This makes it harder than necessary to parse configurations from this repo for vendoring sources. Please let's have some trivial parseable files that contains the URL's and expected output names.

fosslinux:
Certainly a TODO. As some sources are also in scripts in sysc I would like this configuration to be unified.

PR #207, opened 2022-09-29
Change up how sources are defined

fosslinux:
Currently we have sources defined in a very odd way.

- They are defined in the Python get scripts, checksums are in a separate file, and they are also defined in sysc separately.
- Tarballs have to be named to be the same as the package that they are a part of (which is a bit silly).
- Multiple sources isn't supported well.
- Non-tarball sources isn't supported at all.
- If the name of the directory is not the name of the package it needs to be manually specified.

With this system all of these problems are fixed and the sources are defined in a more specced way. This can be interpreted easily by any further tools and allows for significantly more flexibility.


sysa/checksum-transcriber.c
```diff
@@ -0,0 +1,68 @@
...
+	int i;
+	fgets(line, MAX_STRING, in);
+	while (strlen(line) != 0) {
+		// Split each line into tokens
+		tokens = calloc(MAX_TOKENS, sizeof(char*));
+		i = 0;
+		while (i < MAX_TOKENS) {
+			tokens[i] = line;
+			new_line = strchr(line, ' ');
+			// Occurs when there is only two tokens
```
stikonas, line 43:
```suggestion
			// Occurs when there are only two tokens
```

sysa/checksum-transcriber.c
```diff
@@ -0,0 +1,68 @@
...
+			if (new_line == NULL) break;
+			line = new_line;
+			line[0] = '\0';
+			line = line + 1;
+			i = i + 1;
```
stikonas, line 48:
```suggestion
			line += 1;
			i += 1;
```

sysa/checksum-transcriber.c
```diff
@@ -0,0 +1,68 @@
...
+		}
+		line = strchr(line, '\n');
+		line[0] = '\0';
+		// Get checksum and filename
+		checksum = tokens[1];
+		if (tokens[2] != NULL) {
+			filename = tokens[2];
+		} else {
+			filename = strrchr(tokens[0], '/');
+			filename = filename + 1;
```
stikonas, line 58:
```suggestion
			filename += 1;
```

sysa/checksum-transcriber.c
```diff
@@ -0,0 +1,68 @@
...
+		}
+		// Put it all together
+		fputs(checksum, out);
+		fputs("  ", out);
+		// TODO: read env variable (not sure how to do using M2libc)
```
stikonas, line 63: I think you need to get it from `int main(int argc, char** argv, char** envp)`

sysa/run.kaem
```diff
@@ -9,6 +9,10 @@
 
 set -ex
 
+# checksum-transcriber utility
+M2LIBC_PATH=/M2libc
+M2-Mesoplanet --architecture ${ARCH} -f checksum-transcriber.c -o ${bindir}/checksum-transcriber
```
stikonas, line 14: Should we check checksum of `checksum-transcriber`?

sysa/checksum-transcriber.c
```diff
@@ -0,0 +1,68 @@
...
+		fputs("/sysa/distfiles/", out);
+		fputs(filename, out);
+		fputc('\n', out);
+		line = calloc(MAX_STRING, sizeof(char));
+		require(line != NULL, "Failed allocating string");
+	}
+
+	// Clean up
```
stikonas, line 71: I wonder if we can also clean up `malloc`ed memory. Or is that too hard?

.reuse/dep5
```diff
@@ -9,6 +9,6 @@ Source: https://github.com/fosslinux/live-bootstrap
 # Copyright: $YEAR $NAME <$CONTACT>
```
stikonas, line 9: Typo in commit message:
- This is also used to make sources avaliable and extract sources.
should be
- This is also used to make sources available and extract sources.
Also, do we need a keyword to close https://github.com/fosslinux/live-bootstrap/issues/188 ?

stikonas (changes requested):
Looks good in general, mostly changes in checksum-transcriber.c

sysa/checksum-transcriber.c
fosslinux: free is NOP in M2libc anyway. But doesn't hurt anything to clean up

stikonas: Yes, I saw that free is NOP. Though that might be something to rethink for UEFI.

stikonas: Anyway, if it looks complicated to free everything then don't bother, we might have to go with cleanup in M2libc approach anyway, as other software (M2-Planet, hex2, M1, etc) doesn't clean up well either.

fosslinux: All done now.

stikonas (approved):

PR #211, opened 2022-10-20
Update to zlib-1.2.13 and fix its download url.

stikonas:


stikonas:
@fosslinux I'll merge this given this is a fairly critical fix. We can address review comments (if any) later.

fosslinux:
No problems w this.

PR #217, opened 2022-11-27
Zero terminate strings in wstrcpy. Resolves #216.

### LINKED ISSUE #216: Patch with wstrcpy for heirloom-devtools lex should zero terminate string
The wide character patch for lex in heirloom-devtools-070525 does not zero terminate strings. It only works if the destination buffer is zero initialized, but that is not necessarily the case. On the [fiwix kernel](https://github.com/mikaku/Fiwix), the buffer gets non-zero data from the process stack, so the string is not properly terminated, resulting in bad lexer output and a compilation error.

rick-masters:


fosslinux (approved):

fosslinux:
Thanks for the patch!

PR #221, opened 2022-12-09
Ignore checksum for ca-certificates package due to expiring certs.

### LINKED ISSUE #220: ca-certificates fails due to expired certificates
When a certificate expires in the ca-certificates package that certificate is excluded from the package which results in a different package checksum, which fails the build.

I have a PR forthcoming which uses a special keyword IGNORECHECKSUM in place of the checksum in SHA256SUMS.pkgs which fixes this in the least invasive way I can think of. This would need to be replaced every time checksums are regenerated. Perhaps someone has a better idea...

rick-masters:
Fixes #220.

### DIFF
--- a/sysa/SHA256SUMS.pkgs
+++ b/sysa/SHA256SUMS.pkgs
@@ -34,7 +34,7 @@ ae452f08ad9f7ab01f2a9bd882170e71ee6200a06f044f4aadb446cc68700016  bison-3.4.1_1.
 becc8ea1f24c531c35b520ce8bac9e8bf6e30e362fe77bf7833aed9d28af0ae6  bison-3.4.1_2.tar.bz2
 a3266333620a13f4a4c8f4f9819dc6f1fcded7f74bcb9fbf46518f3b160cf930  bison-3.4.2_0.tar.bz2
 09bef4ec506ccafc42ad925f9c038af9885cd8ebca294cf042e63a7d39e91bb5  bzip2-1.0.8_0.tar.bz2
-c484b98e580d7ef5619dc2da2318d9f1e69360a882f64aa15e0fdde0184c45fb  ca-certificates-3.78_0.tar.bz2
+IGNORECHECKSUM  ca-certificates-3.78_0.tar.bz2
 c4d59e759f7349b96f74e62d867e419f3d85fbb81eaf5f79f996abf48f0d3170  coreutils-5.0_0.tar.bz2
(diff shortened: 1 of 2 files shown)


sysa/SHA256SUMS.pkgs
```diff
@@ -34,7 +34,7 @@ ae452f08ad9f7ab01f2a9bd882170e71ee6200a06f044f4aadb446cc68700016  bison-3.4.1_1.
 becc8ea1f24c531c35b520ce8bac9e8bf6e30e362fe77bf7833aed9d28af0ae6  bison-3.4.1_2.tar.bz2
 a3266333620a13f4a4c8f4f9819dc6f1fcded7f74bcb9fbf46518f3b160cf930  bison-3.4.2_0.tar.bz2
 09bef4ec506ccafc42ad925f9c038af9885cd8ebca294cf042e63a7d39e91bb5  bzip2-1.0.8_0.tar.bz2
-c484b98e580d7ef5619dc2da2318d9f1e69360a882f64aa15e0fdde0184c45fb  ca-certificates-3.78_0.tar.bz2
+IGNORECHECKSUM  ca-certificates-3.78_0.tar.bz2
```
nanonyme, line 37: This looks not nice. It prevents anyone pregenerating offline cache from using this file as input. Sounds instead like something that distro should be patching.

nanonyme: The helpers change seems more or less safe.

nanonyme: Never mind. I get your point now. IMHO regardless this file should not be patched. It's important for verifiable builds that everything is checked. Clearly the problem here is the inherent non-determinism in ca-certificates which should probably be patched out, not the checksum itself.

doraskayo:
It would most likely be best to keep expired certificates in the CA bundle for our use case. I think it can be achieved by patching this out:
https://github.com/curl/curl/blob/2bc1d775f510196154283374284f98d3eae03544/scripts/mk-ca-bundle.pl#L585-L593

nanonyme:
Well, that would also change the SHA256 once but at least it would stop being changing over time.

fosslinux:
Yes, that does not matter, we change SHA256's one-off all the time. It matters if those will not match in the future (non-reproducible).

fosslinux:
Closing in favour of approach in #269.

PR #225, opened 2023-01-13
Add a source manifest

doraskayo:

### DIFF
--- a/lib/sysgeneral.py
+++ b/lib/sysgeneral.py
@@ -96,14 +96,28 @@ this script the next time")
                 raise Exception("Download failed.")
         return abs_file_name
 
-    def get_packages(self):
+    def get_packages(self, source_manifest):
         """Prepare remaining sources"""
+        for line in source_manifest.split("\n"):
+            line = line.strip().split(" ")
+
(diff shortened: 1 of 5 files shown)


lib/sysgeneral.py
```diff
@@ -96,14 +96,31 @@ def download_file(self, url, directory, file_name):
                 raise Exception("Download failed.")
         return abs_file_name
 
-    def get_packages(self):
+    def get_packages(self, source_manifest):
         """Prepare remaining sources"""
+
+        for line in source_manifest.split("\n"):
+            line = line.strip().split(" ")
```
stikonas, line 103: Maybe just `split()` which splits on any whitespace

doraskayo: This line is mostly a copy from here: https://github.com/fosslinux/live-bootstrap/pull/225/files#diff-8001649acef5aab633678597aa7db1a43b84a9c9209a8849cf5a8eece3859045R124

I had assumed it was intentionally only expected a newlines.

Would you like me to add a commit to modify both to use `split()`?

stikonas (approved):

PR #227, opened 2023-01-19
parts.rst: add more verbose description of stage0-posix.

stikonas:


parts.rst
```diff
@@ -7,43 +7,131 @@
 
 .. SPDX-License-Identifier: CC-BY-SA-4.0
 
-stage0-posix
-============
+bootstrap-seeds
+===============
+
+This is where it all begins. We start with our ``hex0-seed`` and ``kaem-optional-seed``. First, we use those seeds to rebuild themselves.
```
fosslinux, line 13: can we note these two are raw binary seeds (perhaps define seed) just for clarity?

parts.rst
```diff
@@ -7,43 +7,131 @@
...
+        39D3            ; cmp_ebx,edx                 # Check if we are done
+        0F84 %b         ; je %loop_options_done       # We are done
+        83EB 02         ; sub_rbx, !2                 # --options
+
+hex2
+====
+
+``hex2`` is our final hex language that adds support for labels of arbitrary length. It also allows accessing them via 8, 16, 32-bit relative addresses (``!``, ``@``, ``%``) and via 16-bit or 32-bit (``$``, ``&``) absolute addresses:
+
+.. code:: scheme
```
fosslinux, line 57: why scheme? (here and others)

parts.rst
```diff
@@ -7,43 +7,131 @@
...
+
+    :loop_options
+        39D3                  ; cmp_ebx,edx                        # Check if we are done
+        74 !loop_options_done ; je8 !loop_options_done             # We are done
+        83EB 02               ; sub_ebx, !2                        # --options
+
+catm
+====
+
+``catm`` allows concatenating files with ``catm output_file input1 input2 ... inputN``. This allows us to split shared code into separate files. We will first use it to append ELF header to ``.hex2`` files. Before this step ELF header had to be included in the source file itself.
```
fosslinux, line 67: Sounds like split shared code into separate files means that's what catm can do. Perhaps something along the lines of distributed reused files separately?

Also, "the ELF header"

parts.rst
```diff
@@ -7,43 +7,131 @@
...
+
+M0
+==
+
+The ``M0`` assembly language is the simplest assembly language you can create that enables the creation of real world programs with practical application. It includes only a single keyword: ``DEFINE`` and leverages the language properties of ``hex2`` along with extending the behavior to populate immediate values of various sizes and formats.
```
fosslinux, line 72: Quite subjective first sentence, not sure if I'd describe M0 as really having practical application haha

parts.rst
```diff
@@ -7,43 +7,131 @@
...
+
+    :loop_options
+        cmp_ebx,edx                         # Check if we are done
+        je %loop_options_done               # We are done
+        sub_ebx, %2                         # --options
+
+cc_x86
+======
+
+The ``cc_x86`` implements a subset of the C language designed in ``M0`` assembly. It is somewhat limited subset of C but complete enough to make it easy to write a real C compiler written in the C subset that ``cc_x86`` supports.
```
fosslinux, line 90: "It is a somewhat..."

I'm not sure M2-Planet is a real C compiler. Perhaps, a more full-featured, or a more usable, or something like that instead of real?

parts.rst
```diff
@@ -7,43 +7,131 @@
...
+
+At this stage we start using `M2libc <https://github.com/oriansj/M2libc/>`_ as our C library. In fact, ``M2libc`` ships two versions of C library. At this stage we use a single-file C library that contains just enough to build ``M2-Planet``.
```
fosslinux, line 92: I'd note heere that there is a full version that is rather well-featured.

parts.rst
```diff
@@ -7,43 +7,131 @@
...
+
+M2-Planet
+=========
+
+This is the only C program that we build with ``cc_x86``. `M2-Planet <https://github.com/oriansj/M2-Planet>`_ supports a larger subset of C than ``cc_x86`` and we are somewhat closer to C89 (it does not implement all C89 features but on the other hand it does have some C99 features). ``M2-Planet`` also includes a very basic preprocessor, so we can use stuff like ``#define``, ``#ifdef``.
+
+``M2-Planet`` supports generating code for various architectures including ``x86``, ``amd64``, ``armv7``, ``aarch64``, ``riscv32`` and ``riscv64``.
```
fosslinux, line 99: Could include somewhere that up until this point, bootstrap has been architecture specific.

parts.rst
```diff
@@ -7,43 +7,131 @@
...
+
+``M2-Planet`` is also capable of using full ``M2libc`` C library that has more features and optimizations compared to bootstrap version of ``M2libc``.
+
+mescc-tools
+===========
+
+Now we build ``blood-elf`` used to generate debug info, C version of ``hex2`` (also called ``hex2``) and C version of ``M0`` called ``M1``. These are more capable than their platform specific hex counterparts and are fully cross-platform. Thus we can now have the whole toolchain written in C.
+
+Then rebuild ``mescc-tools`` again, so that they are fully built with toolchain written in C.
```
fosslinux, line 108: Then we rebuild...

Second phrasing is off, maybe "so all our tools are using the new toolchain written in C"?

parts.rst
```diff
@@ -7,43 +7,131 @@
...
+
+Finally, we build `kaem` which is a more capable version of `kaem-optional` and adds support for variables, environmental variables, conditionals, aliases. It also has various built-ins such as `cd` and `echo`.
```
fosslinux, line 110: conditionals and aliases

parts.rst
```diff
@@ -7,43 +7,131 @@
...
+
+M2-Mesoplanet
+=============
+
+``M2-Mesoplanet`` is a preprocessor that is more capable than ``M2-Planet`` and supports ``#include`` statements. It can also launch compiler, assembler and linker, so we don't need to invoke them
```
fosslinux, line 115: More than simply launching, it also automates the invocations and arguments and such - I'd note that.

fosslinux (commented):
Looks reasonably good, only a few nitpicks :)

parts.rst
stikonas: added `raw binary` but did not define seed (not too sure what would be useful to say)

parts.rst
stikonas: Let's remove it, I think this particular blob was from some stage0 readme/file/wiki.

parts.rst
stikonas: just for syntax highlighting purposes

parts.rst
stikonas: Hmm, we do mention it in the next bit about M2-Planet. Do you want to mention it here too?

parts.rst
stikonas: added

parts.rst
fosslinux: Yes, it is a little confusing when we introduce M2libc as a simple library used "just to build M2-Planet" but then it suddenly becomes more well-featured. A casual reader may not instantly draw the link back to this sentence.

fosslinux (approved):

parts.rst
stikonas: ok, updated.

Issue #184, opened 2022-06-14
automake-1.15.1 failed to build


stikonas:
Probably some race condition

```
automake-1.15.1: compiling source.
make[1]: Entering directory `/usr/src/automake-1.15.1/build/automake-1.15.1'
make[1]: Leaving directory `/usr/src/automake-1.15.1/build/automake-1.15.1'
Updating ./doc/version.texi
  MAKEINFO doc/automake.info
  MAKEINFO doc/automake-history.info
  GEN      bin/automake
  GEN      bin/aclocal
  GEN      t/ax/shell-no-trail-bslash
  GEN      t/ax/cc-no-c-o
  GEN      runtest
  GEN      doc/aclocal.1
  GEN      doc/automake.1
  GEN      lib/Automake/Config.pm
  GEN      doc/aclocal-1.15.1
  GEN      doc/automake-1.15.1
  GEN      doc/amhello-1.0.tar.gz
doc/amhello-1.0.tar.gz: recipe failed.
See file '/usr/src/automake-1.15.1/build/automake-1.15.1/doc/amhello/amhello-output.tmp' for details
make: *** [doc/amhello-1.0.tar.gz] Error 1
```

stikonas:
That is possibly caused by something broken in autotools. I'll be reworking autotools symlinks, so this might be fixed then.

stikonas:
Still happens from time to time.

stikonas:
`amhello-output.tmp` end with

```
  && make  install \
  && make  installcheck \
  && make  uninstall \
  && make  distuninstallcheck_dir="$dc_install_base" \
        distuninstallcheck \
  && chmod -R a-w "$dc_install_base" \
  && ({ \
       (cd ../.. && umask 077 && mkdir "$dc_destdir") \
       && make  DESTDIR="$dc_destdir" install \
       && make  DESTDIR="$dc_destdir" uninstall \
       && make  DESTDIR="$dc_destdir" \
            distuninstallcheck_dir="$dc_destdir" distuninstallcheck; \
      } || { rm -rf "$dc_destdir"; exit 1; }) \
  && rm -rf "$dc_destdir" \
  && make  dist \
  && rm -rf amhello-1.0.tar.gz \
  && make  distcleancheck \
  && cd "$am__cwd" \
  || exit 1
checking for a BSD-compatible install... /usr/bin/install -c
checking whether build environment is sane... yes
[...]

PR #231, opened 2023-01-24
Do not build doc directory of autoconf 2.11 and 2.15.

stikonas:
There is a race condition there that often causes builds to fail.
Fixes: #184


fosslinux (approved):

fosslinux:
Shouldn't commit message say automake?

stikonas:
> Shouldn't commit message say automake?

Oh yes, I'll fix that.

PR #233, opened 2023-01-28
Rework tmpdir & associated a bit.

fosslinux:
- Split out tmpdir logic into a separate entity & add the appropriate arguments and checks.
- sysb can be removed since there is now no associated logic.
- Move disk/etc logic into tmpdir.py.

Also remove minikernel since it is unmaintained and broken (and soon useless).

This was done keeping in mind new kernel setup coming soon.


stikonas (approved):

fosslinux:
Wasn't able to make this break in any other weird ways.

Issue #230, opened 2023-01-24
Support live-bootstrapping only up to sysc_image and resuming sysc separately


nanonyme:
Currently when you use chroot/bwrap method for live-bootstrapping, you need to chroot at the end of sysa. Allowing separate top-level sandboxes for sysa and sysc allows more flexible integration to build systems. Additionally, having separate sandbox for sysc allows even with direct rootfs usage dropping chroot permission from bwrap as you can simply create a new bwrap sandbox.

nanonyme:
Looks like also splitting build like this with bwrap would help in cases like https://edition.cnn.com/2023/01/21/europe/turkey-cancels-swedish-defense-minister-visit-intl/index.html as when we drop the chroot permission, you might be able to run setuid bwrap just fine.

fosslinux:
I like the idea; for chroot/bwrap only, though. This is not appropriate for QEMU/bare metal where we use kexec.

nanonyme:
Right, there are already guards with regards to chroot https://github.com/fosslinux/live-bootstrap/blob/master/sysa/run.sh#L251

stikonas:
Ideally this should be provided as an option (even in chroot and bwrap modes), while we could still use the old chroot method.

nanonyme:
To be honest, in bwrap mode we probably should just always use separate sandboxes. It will already add extra code to add the separate sandboxes and to *drop* chroot permission inside bwrap (it's better to ensure that works since it's not guaranteed to be available for everyone using live-bootstrap) we would closer to triple the current amount of code. For chroot I'm ambivalent whether it buys much to use separate chroots rather than nested since you are root already when you use chroot.

nanonyme:
So in other words
1. Extra option in bootstrap.cfg
2. Using bwrap implies this option is true

nanonyme:
@stikonas I think we *probably* do not want to high-level expose this in chroot method since you *already* have chroot permission so splitting doesn't really help. But with bwrap it allows dropping a cap which may help when no user namespaces are available.

PR #235, opened 2023-02-01
Support building only sysa with chroot mode

nanonyme:
This is intended primarily for external build systemss that directly bind into sysa and sysc rather than using rootfs Python wrapper. Part of https://github.com/fosslinux/live-bootstrap/issues/230

### DIFF
--- a/sysa/run.sh
+++ b/sysa/run.sh
@@ -250,4 +250,6 @@ fi
 # In chroot mode transition directly into System C.
 SYSC=/sysc_image
 sys_transfer "${SYSC}" /sysc gzip patch
-exec chroot "${SYSC}" /init
+if [ "${CHROOT_ONLY_SYSA}" != True ]; then
+    exec chroot "${SYSC}" /init
+fi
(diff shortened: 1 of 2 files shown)


nanonyme:
So this initial case targets https://github.com/fosslinux/live-bootstrap#python-less-build; we can extend rootfs.py bwrap mode to have separate sandboxes (and drop chroot privilege) in further work if desired.

sysa/run.sh
```diff
@@ -250,4 +250,6 @@ fi
 # In chroot mode transition directly into System C.
 SYSC=/sysc_image
 sys_transfer "${SYSC}" /sysc gzip patch
+if [ "${CHROOT_ONLY_SYSA}" != True ]; then
 exec chroot "${SYSC}" /init
```
stikonas, line 254: Can you add some indentation here

stikonas:
Could you also rebase it on top of master. I've pushed pylint fixes (probably caused by pylint upgrade).

sysa/run.sh
nanonyme: Fixed.

stikonas (approved):

PR #238, opened 2023-02-04
Pre-GCC10 changeset

fosslinux:
All of this was originally meant to sit on the GCC10 PR, since most of this is required for GCC10.

However, I decided to separate it out (and add a few harmless fixes on top).

Not bothering to split this into multiple PRs because almost every commit touches a number of checksums.

Changes required for GCC 10:

- binutils
  - Use `i386-unknown-linux-musl` instead of `i386-unknown-linux-gnu` for 2.38
  - Add `$prefix-$cmd` symlinks
  - Don't install to `/usr/$target` (breaks detections in GCC)
  - Remove with-sysroot option - breaks GCC 10 invocation of `ld` (surprised it didn't in GCC 4.x)
- util-linux
  - Don't install to `/usr/$target` (breaks detections in GCC)

Other cleanup:

- `LIBDIR` pre-musl
- TCC (broke sysa with `--repo` when recompiling something in sysa)
- Packaging bug following symlinks to directories (hasn't ever come up, but may..)

I'm pretty sure most of these changes are harmless and could be pushed straight to `master`, but they build up to something a bit more substantial, hence the PR.


stikonas (approved):
Only a couple of typos in commit messages:

obselete -> obsolete
by change -> by chance

PR #240, opened 2023-02-06
Split sysa and sysc when using bwrap

nanonyme:
Using separate bwrap invocations allows dropping chroot permission. Fixes https://github.com/fosslinux/live-bootstrap/issues/230


doraskayo:
Looks correct.

stikonas (approved):
Looks correct to me too.

PR #242, opened 2023-02-11
GCC 10

fosslinux:
nearly done

TODO

- [x] fix up checksums due to toolchain changes (almost all debug symbols have changed due to gcc(?) patch)
- [x] double check qemu & chroot reproducibility
- [x] update `parts.rst`
- [x] reuse

**warning: package checksums are still wrong**


sysc/gcc-10.4.0/gcc-10.4.0.sh
```diff
@@ -0,0 +1,115 @@
...
+
+    # Regenerate crc table in libiberty/crc32.c
+    pushd libiberty
+    sed -n -e '38,65p' crc32.c > crcgen.c
+    gcc -o crcgen crcgen.c
+    head -n 69 crc32.c > crc32.c.new
+    ./crcgen >> crc32.c.new
+    tail -n +138 crc32.c >> crc32.c.new
+    mv crc32.c.new crc32.c
+    popd
```
stikonas, line 67: argh, we need to do this for older GCC too. I think only 4.7.4 is affected but not 4.0.4.

stikonas (commented):
Looks fine in general (once reproducibility issues are fixed).

sysc/gcc-10.4.0/gcc-10.4.0.sh
fosslinux: oh, good point.. i'll do that after this PR though

fosslinux:
checksums should be good now

stikonas (approved):

stikonas (approved):

PR #245, opened 2023-02-12
Commit_fix

### LINKED ISSUE #244: Typo in util-linux
https://github.com/fosslinux/live-bootstrap/blob/4b5b0377184680de08dff87e0e7f4fa909a3bbec/sysa/util-linux-2.19.1/util-linux-2.19.1.sh#L21

Build string refers to `fdsik`. It should be `fdisk` but given that it still seems to work, do we even need it?

tarunsamanta2k20:
Fixes #244

### DIFF
--- a/sysa/util-linux-2.19.1/util-linux-2.19.1.sh
+++ b/sysa/util-linux-2.19.1/util-linux-2.19.1.sh
@@ -18,7 +18,7 @@ src_configure() {
         --disable-libuuid \
         --without-ncurses \
         --enable-static \
-        --enable-static-programs=losetup,mount,umount,fdsik,sfdisk,blkid \
+        --enable-static-programs=losetup,mount,umount,fdisk,sfdisk,blkid \
         --enable-shared=no \
         --disable-wall \
         ac_cv_type_loff_t=yes


stikonas (changes requested):
* You need a better commit message, `Commit_fix` does not give any information to other persons what was fixed. Also it's a good idea to link commit to original issue, you could add `Fixes #244` to the commit message (but on the new line, not in the commit title)
* Most changes in live-bootstrap affect checksums of output products. You need to run live-bootstrap and then record a new checksum in `sysa/SHA256SUMS.pkgs`. I ran live-bootstrap with your change and got
```
make[2]: Leaving directory `/sysa/util-linux-2.19.1/build/util-linux-2.19.1'
make[1]: Leaving directory `/sysa/util-linux-2.19.1/build/util-linux-2.19.1'
util-linux-2.19.1: creating package.
util-linux-2.19.1: checksumming created package.
util-linux-2.19.1_0.tar.bz2: FAILED
sha256sum: WARNING: 1 of 1 computed checksum did NOT match
```
* I think we can just remove the whole `--enable-static-programs` line completely since everything is actually static. @fosslinux do you remember why you added `--enable-static-programs`. By the way, removing the whole line would also change checksum.

fosslinux:
No, i don't really remember any good reason for enable-static-programs - I guess I was probably reading configure and had a bit of a brainfart.

stikonas (changes requested):
You still need to address all 3 bullet points above. We can't merge it like it is now as that will break the runs.

PR #248, opened 2023-02-21
Binutils prerequisites

stikonas:
Mostly moving newer autotools before binutils.

I've kept the latest automake 1.15 in sysc as it needs newer bash. (Perhaps it's still useful to move it immediately after bash to upgrade xz).


stikonas:
Disabled misleading shellcheck warning. Our globs end up in filelist.txt, so should not cause any problems even if files are called --something.

stikonas:
(rebased on top of merged mes-0.24.2)

fosslinux (approved):

PR #250, opened 2023-02-24
Upgrade to binutils 2.24:

stikonas:
* Move libtool 2.2.4 before binutils.
* Remove libtool 1.4.
* Switch from binutils 2.14 to 2.24.


sysa/binutils-2.24/binutils-2.24.sh
```diff
@@ -0,0 +1,103 @@
...
+    rm bfd/doc/bfd.info binutils/doc/binutils.info
+    rm etc/configure.info etc/standards.info
+    rm gas/doc/as.info gprof/gprof.info ld/ld.info
+
+    # Remove pregenerated opcodes files
+    rm opcodes/i386-init.h opcodes/i386-tbl.h
+    rm opcodes/ia64-asmtab.c
+    rm opcodes/z8k-opc.h
+    rm opcodes/aarch64-asm-2.c opcodes/aarch64-opc-2.c opcodes/aarch64-dis-2.c
+
```
fosslinux, line 56: There is significantly more than this - we are missing this from binutils 2.38 too.

`grep 'MACHINE GENERATED' -r opcodes`

I think the vast majority is safe to just delete though, but lets not leave it behind in the tree

fosslinux (commented):
only one small issue

sysa/binutils-2.24/binutils-2.24.sh
stikonas: Maybe it's worth moving findutils back before binutils to delete them? Or should I just list them all?

fosslinux: Your call, or just (i think we have xargs at this point)

`grep -l 'MACHINE GENERATED' -r opcodes | xargs rm`

stikonas: xargs is part of rm but I can use rm $() for this

stikonas: Strange, diffoscope now shows
```
│ │ --rw-r--r--   0        0        0        0 1970-01-01 00:00:00.000000 ./usr/share/man/man1/ld.1
│ │ +-rw-r--r--   0        0        0   113875 1970-01-01 00:00:00.000000 ./usr/share/man/man1/ld.1
```
We are missing ld.1 after I remove those generated opcodes. In fact I think we should remove all of them...

PR #251, opened 2023-03-01
Binutils 2.30

stikonas:


sysa/bc-1.07.1/bc-1.07.1.sh
```diff
@@ -0,0 +1,21 @@
+# SPDX-FileCopyrightText: Andrius Štikonas <andrius@stikonas.eu>
```
fosslinux, line 1: copyright year?

sysa/ed-1.4/ed-1.4.sh
```diff
@@ -0,0 +1,16 @@
+# SPDX-FileCopyrightText: Andrius Štikonas <andrius@stikonas.eu>
```
fosslinux, line 1: same for copyright year

sysa/run.sh
```diff
@@ -193,58 +193,6 @@ build autoconf-2.69
...
-    go_sysb
-fi
-
-# In chroot mode transition directly into System C.
-SYSC=/sysc_image
-sys_transfer "${SYSC}" /sysc gzip patch
-if [ "${CHROOT_ONLY_SYSA}" != True ]; then
-    exec chroot "${SYSC}" /init
-fi
+exec env -i PATH="${PATH}" PREFIX="${PREFIX}" LIBDIR="${LIBDIR}" SOURCES="${SOURCES}" DISTFILES="${DISTFILES}" DESTDIR="${DESTDIR}" SRCDIR="${SRCDIR}" bash run2.sh
```
fosslinux, line 198: Hmm, if we are doing this in sysa now, it might make more sense to do the `.env` stuff that we use in sysc in sysa as well.

fosslinux: This doesn't have to be in this PR though

fosslinux (approved):
Generally this looks fine to me. Just the copyright years before merge, but let's do the .env stuff soon too.

sysa/run.sh
stikonas: Yes, I need to redo it for sysa but that PR was already quite large.

PR #253, opened 2023-03-01
Implement kernel bootstrap up to building Linux.

### LINKED ISSUE #252: Integrate kernel bootstrapping from boot2now project
Please integration kernel bootstrapping from the boot2now project. Kernel bootstrapping works up to building Linux and has been integrated with live-bootstrap with a new `--kernel-bootstrap` option.

There is one last step remaining to complete the kernel bootstrap - a kexec of Linux from the Fiwix kernel. However, it may make sense to integrate with what works now in order to keep the integration up to date. Integration now would allow it to be tested and fixed if necessary when updates are made to live-bootstrap.

A PR request will be added to this issue. This PR is large and can be considered a first draft. I expect there will be substantial questions and suggestions which may lead to either updating the PR or creating a new one. I'd still like to improve the performance of builder-hex0 loading files at the beginning, improve the documentation, and do more testing.

The PR comes from the `kernel-bootstrap-v1` branch of the http://github.com/rick-masters/live-bootstrap repo. So, testing of the integration can be done directly from that repo/branch.

rick-masters:
Resolves #252


parts.rst
```diff
@@ -34,6 +38,12 @@ In the first steps we use initial ``hex0`` binary seed to rebuild ``kaem-optiona
 
 ``hex0`` can be approximated with: ``sed 's/[;#].*$//g' $input_file | xxd -r -p > $output_file``
 
+
+builder-hex0 (kernel bootstrap)
+===============================
+If the kernel-bootstrap option is enabled then the ``builder-hex0`` kernel boots from a hard drive and loads an enormous shell script which embeds files (loaded with the ``src`` command) and the initial commands to build ``hex0-seed``, ``kaem-optional-seed``, and the command which launches stage0-posix using ``kaem-optional-seed`` and the stage0-posix launch script ``kaem.x86``. Builder-hex0 is written in hex0 and can be compiled with the ``hex0-seed``, ``sed``, the tiny ``builder-hex0-mini`` boot kernel or it can build itself.
```
fosslinux, line 44: can be compiled with any one of ....

Initially I thought that this meant you needed all of them to build it, and quickly realized that made no sense.

sysa.py
```diff
@@ -100,3 +106,125 @@ def stage0_posix(self):
...
+
+        for filepath in files:
+            if ".git" in filepath:
+                continue
+            self.output_file(srcfs, filepath)
+
+        os.chdir(self.tmp_dir)
+        shutil.move(os.path.join('sysa','stage0-posix'), '.')
+        shutil.move(os.path.join('sysa', 'distfiles'), '.')
+        run('tar', '-c', '-z', '-p',
```
fosslinux, line 175: Is there a reason for tarring vs just appending every file? If it's required then so be it for now, but I'm not a huge fan of requiring tar on the host system.

sysa/fiwix-1.4.0-lb1/sources
```diff
@@ -0,0 +1 @@
+https://github.com/rick-masters/Fiwix/releases/download/v1.4.0-lb1/fiwix-1.4.0-lb1.tar.gz ff57d1fbf0547a27dec4d7146eb9c17d01158ca5886de54e5b9bac2bec473ef3
```
fosslinux, line 1: In the future, do you think the patchset will be small enough that we be able to either use an upstream release of Fiwix or apply any remaining patches within live-bootstrap? I'm don't really love using live-bootstrap specific forks, unless the number of patches is prohibitive.

sysa/fiwix-1.4.0-lb1/fiwix-1.4.0-lb1.kaem
```diff
@@ -0,0 +1,554 @@
+#!/bin/sh
+
+# SPDX-FileCopyrightText: 2021-22 fosslinux <fosslinux@aussies.space>
+# SPDX-FileCopyrightText: 2021 Andrius Štikonas <andrius@stikonas.eu>
```
fosslinux, line 4: this should be you ;)

sysa/fiwix-1.4.0-lb1/fiwix-1.4.0-lb1.kaem
```diff
@@ -0,0 +1,554 @@
...
+	lib/printk.o \
+	-o /boot/fiwix
+
+# Checksums
+if match x${UPDATE_CHECKSUMS} xTrue; then
+    sha256sum /boot/fiwix
+    #sha256sum -o ${pkg}.checksums \
+    #    /boot/fiwix
+
+    #cp ${pkg}.checksums ${srcdir}
```
fosslinux, line 551: Fine for draft, just a reminder to fix before merge

sysa/lwext4-1.0.0-lb1/lwext4-1.0.0-lb1.kaem
```diff
@@ -0,0 +1,47 @@
...
+tcc -m32 -march=i386 -std=c89 -I../include -I../build_generic/include -I../../tcc/tcc-0.9.27/include -c ext4_super.c
+tcc -m32 -march=i386 -std=c89 -I../include -I../build_generic/include -I../../tcc/tcc-0.9.27/include -c ext4_trans.c
+tcc -m32 -march=i386 -std=c89 -I../include -I../build_generic/include -I../../tcc/tcc-0.9.27/include -c ext4_xattr.c
+
+catm make_fiwix_initrd.c ../../../files/make_fiwix_initrd.c
+tcc -m32 -march=i386 -std=c89 -I../include -I../build_generic/include -I../../tcc/tcc-0.9.27/include -DVERSION="\"1.0\"" -c make_fiwix_initrd.c
+
+tcc -m32 -o make_fiwix_initrd ext4.o ext4_balloc.o ext4_bcache.o ext4_bitmap.o ext4_block_group.o ext4_blockdev.o ext4_crc32.o ext4_debug.o ext4_dir.o ext4_dir_idx.o ext4_extent.o ext4_fs.o ext4_hash.o ext4_ialloc.o ext4_inode.o ext4_journal.o ext4_mbr.o ext4_mkfs.o ext4_super.o ext4_trans.o ext4_xattr.o ../blockdev/linux/file_dev.o make_fiwix_initrd.o
+
+./make_fiwix_initrd
```
fosslinux, line 47: I would prefer that this is run outside of the build script in `run.kaem` - build scripts should really only be doing compiling & installing.

sysa/mes-0.24.2/simple-patches/mes-0.24.1.kaem-base-address.after
```diff
@@ -0,0 +1 @@
+${mescc_scm} -- --base-address 0x08048000 -L
```
fosslinux, line 1: Can you add documentation for `simple-patches` subdir to `DEVEL.md`.

fosslinux (changes requested):
This is a first-pass review, there's some things I haven't really looked at in detail yet but there's a few larger architectural things;

1. I would like for `--kernel-bootstrap` to be the default, when QEMU is used. This is meant to be the default pathway. Then, if `--kernel` is given, *then* we can not use the fiwix/builder-hex0 stuff.
2. I apologise for not communicating this one earlier, as this has been on my mind for a while with this. I'd like, when running under Fiwix, for an initramfs *not* to be used (or, at least, used to the minimal extent possible - but from my relatively brief look over this it doesn't appear required). At minimum, it'd be great if all the packages could be build on-disk rather than in-ram (initramfs). This will allow live-bootstrap to be build on much lower memory machines and allow for sysa to be expanded without issue. builder-hex0 does look like it can write to/read from disk, but I'm not sure if that extends to the creation of an EXT filesystem on a disk?
3. This one is more of a question - what's the rationale for using builder-hex0 to start, instead of say, builder-hex0-mini? (it probably doesn't really matter, because both are hex0 anyway, but just wanted to hear your thoughts).

Thank you very much for this PR, I'm excited to get this in :tada:

fosslinux:
Also, with your comment in #252, will this currently fail or do something else when transferring from fiwix -> linux?

Other thing I just thought of, architecturally I think it would be more consistent to split sysa into sysa and sysb, where sysa is builder-hex0 and sysb is fiwix, and then the current sysb  can be removed, since we have guranteed hard disk support in fiwix.

parts.rst
```diff
@@ -665,6 +704,14 @@ sysb is purely a transition to sysc, allowing binaries from sysa to get onto a
...
 It populates device nodes, mounts sysc, copies over data, and executes sysc.
 
+curl 7.83.0
+===========
+Curl is built in sysc because Linux must be running with support for threads.
+Curl requires musl 1.2.3 with thread support which was built at the end of sysa.
+Curl is built first in sysc so the rest of the packages can be downloaded.
+Note that the tar file for curl itself was copied over from sysa because
+curl is not yet available to download it.
+
```
stikonas, line 714: Would it not be simpler to build curl at the end of sysa then? Obviously it wouldn't be able to run on fiwix, but then we can transfer curl binary (together with all other binaries that we already do) rather than special casing tarball transfer.

rick-masters: I'm rebuilding musl-1.2.3 with threads at the end of sysa but that version of musl-1.2.3 cannot be used because Fiwix does not support it. Unfortunately if a curl build is attempted `configure` will create test programs for threads that pick up the new musl and end up not working. Worse, they hang the build.

sysa/fiwix-1.4.0-lb1/sources
rick-masters: It it my goal and hope that we can use an official release and then patch it. Unfortunately right now there are too many complicated patches. We only have simple-patch at that stage and it would be a lot of work to construct patches that way.

sysa/lwext4-1.0.0-lb1/lwext4-1.0.0-lb1.kaem
rick-masters: This will be resolved in the next update.

sysa/mes-0.24.2/simple-patches/mes-0.24.1.kaem-base-address.after
rick-masters: This will be resolved in the next update.

parts.rst
stikonas: Yes, makes sense. Oh well...

sysa.py
rick-masters: The usage of tar will be removed in the next update.

rick-masters:
@fosslinux 
For your other questions:
1. I will work on making removing `--kernel-bootstrap` and make it the default instead.
2. I will explore writing the ext2 image to disk instead of using a ram drive. This would mean overwriting the builder-hex0 srcfs disk image. One issue is that I believe it may exceed the size of a hard drive supported by a simple BIOS using Cylinder/Head/Sector calculations, at least with the current default qemu disk geometry. It might require LBA but I'm not sure.
3. I think it is a reasonable goal to start with something small like builder-hex0-mini that complies builder-hex0. The problem with the current design of builder-hex0-mini is that it writes the compiled binary to the start of disk and depends on a reboot to activate it. As with the transition to Fiwix I think a reboot can be avoided. This would take some work. I'd postpone this until after the initial integration.
4. There is currently no transition from Fiwix to Linux so the bootstrap stops soon after Linux is compiled. That may be a reason to postpone making `--kernel-bootstrap` the default initially until I can complete the Fiwix to Linux boot transition.

It is not very clear to me how a redesign with Fiwix as sysb would work exactly. Sysb is currently used for booting Linux (I think) so I don't see how that changes. (To be honest, I'm not sure I want to sign up for changing the method of booting Linux.) I think I'd need some more details to understand what you are suggesting.

It might be a while until I submit the next updated PR. I'm taking a breather to work at a more leisurely pace for a little while.

stikonas:
> @fosslinux For your other questions:
> 
>     2. I will explore writing the ext2 image to disk instead of using a ram drive. This would mean overwriting the builder-hex0 srcfs disk image. One issue is that I believe it may exceed the size of a hard drive supported by a simple BIOS using Cylinder/Head/Sector calculations, at least with the current default qemu disk geometry. It might require LBA but I'm not sure.

Perhaps it might be simpler to do something like this:
* builder-hex0 boots fiwix with ram drive.
* fiwix then has a hard drive support, so we then populate the hard drive and transition into another fwix.

This would add another kexec like transition but perhaps code can mostly be shared with fiwix to linux transition?

sysa/fiwix-1.4.0-lb1/sources
fosslinux: Sounds good to me. Happy with this as-is in that case :)

fosslinux:
> I will explore writing the ext2 image to disk instead of using a ram drive. This would mean overwriting the builder-hex0 srcfs disk image. One issue is that I believe it may exceed the size of a hard drive supported by a simple BIOS using Cylinder/Head/Sector calculations, at least with the current default qemu disk geometry. It might require LBA but I'm not sure.

Sounds good. If you think that @stikonas method would be easier, or if two separate disks are used for builder-hex0 and fiwix, that's alright with me.

> I think it is a reasonable goal to start with something small like builder-hex0-mini that complies builder-hex0. The problem with the current design of builder-hex0-mini is that it writes the compiled binary to the start of disk and depends on a reboot to activate it. As with the transition to Fiwix I think a reboot can be avoided. This would take some work. I'd postpone this until after the initial integration.

No worries about this for the time being, then - this can come later, as you said.

> There is currently no transition from Fiwix to Linux so the bootstrap stops soon after Linux is compiled. That may be a reason to postpone making --kernel-bootstrap the default initially until I can complete the Fiwix to Linux boot transition.

Ah, ok. In my opinion, let's still make `--kernel-bootstrap` the default, but document/add warnings that `--kernel` should still be used for a full build. Happy to hear other thoughts though.

> It might be a while until I submit the next updated PR. I'm taking a breather to work at a more leisurely pace for a little while.

That is absolutely fine! It is good to spend some time for ourselves.

rick-masters:
@stikonas 

`fiwix then has a hard drive support, so we then populate the hard drive and transition into another fwix.`

I'm not sure this question matters yet. builder-hex0/BIOS may be able to write to a larger hard disk by using a bigger disk geometry. I just need to look into it. If not, maybe your Linux changes will save enough space. We need to save approx. 140MB to keep under the limit of the current geometry. (Which is 1024 cylinders * 32 heads * 63 sectors * 512 bytes/sector)

@fosslinux 
Note that eliminating a Fiwix ram drive does not reduce overall memory requirements. Currently builder-hex0 uses 3GB which is more than Fiwix. The usage is approx. 650MB for live-bootstrap distfiles and builds, 1152MB for the Fiwix ext2 image, 670MB for running processes, and 604MB for suspended processes. Even if the ext2 image is copied to disk it needs to be created in memory first (with the current design of builder-hex0.)

The vast majority of the 1.2GB for processes is for mes. I could eliminate half (the memory for suspended processes) by implementing virtual memory but I'm reluctant to add that complexity. Long term really getting memory down means redesigning or replacing mes, IMHO.

That said, there is a significant benefit to moving away from a Fiwix ram disk. My modification to Fiwix to support large ram drives was difficult/complex/messy and I'd rather not try to upstream it:
https://github.com/rick-masters/Fiwix/commit/747a76335a30d41ada7a7803f5ef917068199843

So, I'll continue to look at how to boot Fiwix off a hard drive. For the transition to Linux, I expect sysb/run.sh would need to be modified to do one of: format the first drive (overwriting sysa), format a second partition on the first drive,  or format a (or use a pre-formatted) second drive.

fosslinux:
>  Note that eliminating a Fiwix ram drive does not reduce overall memory requirements. Currently builder-hex0 uses 3GB which is more than Fiwix. The usage is approx. 650MB for live-bootstrap distfiles and builds, 1152MB for the Fiwix ext2 image, 670MB for running processes, and 604MB for suspended processes. Even if the ext2 image is copied to disk it needs to be created in memory first (with the current design of builder-hex0.)

This is not particularly problematic. If sysa is expanded or more packages are added, then the only added cost to RAM is distfiles & possibly the ext2 image. However, in the current state, we also need to have disk space available for the compilation process - this is where compiling large packages such as the Linux kernel can become very problematic. With Fiwix it seems we should be able to eliminate this problem.

> So, I'll continue to look at how to boot Fiwix off a hard drive. For the transition to Linux, I expect sysb/run.sh would need to be modified to do one of: format the first drive (overwriting sysa), format a second partition on the first drive, or format a (or use a pre-formatted) second drive.

If you can handle the Fiwix side, I'd be happy to do this bit.

rick-masters:
@fosslinux 
`With Fiwix it seems we should be able to eliminate this problem.`

I'm not sure how. If we need more space for compilation then the ext2 image needs to be expanded and since that is created in memory with builder-hex0, it expands the memory requirements. builder-hex0 is near usable memory limits already which is 3GB without implementing PAE.
Perhaps you are suggesting Fiwix could expand the ext2 file system once on disk or format another partition or disk?

sysa/kexec-fiwix/src/kexec-fiwix.c
```diff
@@ -0,0 +1,217 @@
...
+	*((unsigned int *) &trampoline[11]) = INITRD_MB * 1024 * 1024;
+	*((unsigned int *) &trampoline[19])  = magic;
+	*((unsigned int *) &trampoline[24])  = multiboot_info_num;
+	*((unsigned int *) &trampoline[29])  = e_entry;
+
+	memcpy((void *)0x4000, trampoline, sizeof(trampoline));
+
+	printf("kexec-fiwix: jumping to trampoline...\n");
+	__asm__ __volatile__ (
+		".byte 0xEA, 0x00, 0x40, 0x00, 0x00, 0x08, 0x00 \n\t"
```
stikonas, line 215: Can you not use ljmp and let tcc convert it to machine code here instead of writing machine code directly?

rick-masters: This will be resolved in the next update.

fosslinux:
superseded by #272

PR #255, opened 2023-03-04
Reduce the size of linux kernel to free up some space.

stikonas:


fosslinux (approved):
note to self: this wont change any checksums because package gen for linux is intentionally broken

stikonas:
> note to self: this wont change any checksums because package gen for linux is intentionally broken

Given that I removed object files after linking but before installation in earlier PR, we might have enough space to create package.

I think max memory usage is now at the kernel linking stage and goes down afterwards.

PR #266, opened 2023-03-14
Fix checksum of gettext package

pder:
The checksum of this package changed due to commit 76a5639b30f786fcaf22c0 "Strip created binaries & libraries"


fosslinux:
Thanks! I was about to push this change but you just beat me to it :)

Issue #209, opened 2022-10-19
https://zlib.net/zlib-1.2.12.tar.xz returns 404


schierlm:
When downloading zlib sources, it fails as https://zlib.net/zlib-1.2.12.tar.xz returns 404.
Older versions of zlib (https://zlib.net/fossils/) seem to be only available as tar.gz, not tar.xz.

But in my opinion worse: The downloading just stops at that point (I did not notice any error message) and it tries to bootstrap anyway with only 70 files in `sysa/distfiles` (from line count of `sources` files there should be 132).

stikonas:
Let's switch to fossils UR (I can upgrade to 1.2.13 at the same time ). Though that wouldn't yet fix lack of error message.

stikonas:
Fix for immediate problem https://github.com/fosslinux/live-bootstrap/pull/211 but not for download failure.

schierlm:
Thanks for fixing. To be honest, I am no longer sure about the thing with the error message. Seems I ran it without `--external-sources`, so it only downloaded stuff needed for sysa, not sysc (which does not include zlib). and did not notice as the chroot was created without `unshare -net` and just downloaded the files inside instead.

stikonas:
I'll test it then in a couple of days if I have time and if there is not error then we can close this.

PR #270, opened 2023-03-19
Add `-f` flag to curl download commands

eduardosm:
By default, curls downloads the HTML error page and exits with code 0 when a server replies with an HTTP error code (e.g., 404), causing a SHA256 mismatch afterwards.

Adding the `-f` flag makes curl exit with non-zero error code and print an error like "The requested URL returned error: 404", making it a bit easier to diagnose distfile download issues.

### DIFF
--- a/download-distfiles.sh
+++ b/download-distfiles.sh
@@ -15,7 +15,7 @@ download_source() {
     local dest_path="${distfiles}/${fname}"
     if ! [ -e "${dest_path}" ]; then
         echo "Downloading ${fname}"
-        curl -L "${url}" --output "${dest_path}"
+        curl --fail --location "${url}" --output "${dest_path}"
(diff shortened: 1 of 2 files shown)


stikonas (commented):
Looks good but can you use long option --fail.

Using long options in scripts makes them more readable (I know we don't always follow this rule here... But let's not make problem bigger). 

And it would also be good to fix -L to --location

eduardosm:
I agree, replaced `-fL` with `--fail --location`.

stikonas (approved):

schierlm:
Hmm, probably that was the same issue I saw in https://github.com/fosslinux/live-bootstrap/issues/209 but was not able to reproduce. Thanks for fixing :)

PR #271, opened 2023-03-19
Update curl to 7.88.1

eduardosm:
Patches have been regenerated to update hunk line numbers


stikonas (approved):
Thanks for this PR.

I briefly checked tarball for potential pre-generated files but didn't spot anything new.

stikonas:
@rick-masters This actually conflicts a bit with your PR
I don't think it would break Fiwix, but it would cause some merge conflicts. So I would like to have your opinion before mering this. Do we merge it or hold it off till your PR is merged.

sysa/curl-7.88.1/sources
```diff
@@ -0,0 +1 @@
+https://curl.se/download/curl-7.88.1.tar.bz2 8224b45cce12abde039c12dc0711b7ea85b104b9ad534d6e4c5b4e188a61c907
```
fosslinux, line 1: Can we change this back to sourceforge for now? If we keep it like this no `--external-sources` is broken. @stikonas thoughts?

stikonas: That's sysa, why would it break no external sources?

As for sysc, is http broken on curl.se?

rick-masters:
@stikonas 

Regarding the conflict with kernel bootstrap PR, I will merge it so go ahead. I'll also test it soon and if there is an unforeseen problem I'll let you know.

sysa/curl-7.88.1/sources
stikonas: @fosslinux ok, I've tested and it is indeed broken.
So I'm unsure about what to do... But we know that it will get broken again if we use sourceforge link.

Anyway, longer term this might not be too important as @rick-masters PR transfers curl tarball from fiwix system to linux system and builds first builds of curl at the very beginning of sysa (and then rebuilds with openssl later). But that means we wont be doing any downloading of curl in sysc.

fosslinux: Yeah, I'm not sure which is worse, having `--external-sources` required or knowing it will break given time again..

stikonas: which effectively means we can't use curl that only supports HTTP. Anyway, this will be reworked by rick-masters, so perhaps we should just apply his workaround?

I.e. move first curl stage to sysc (but transfer tarball from sysa, so effectively we use predownloaded tarball).

eduardosm: > I.e. move first curl stage to sysc (but transfer tarball from sysa, so effectively we use predownloaded tarball).

Let's try that. I did not have time today to fully test this locally, so I pushed and I will check tomorrow the result of GH Actions.

stikonas: ok, if it passes, I think we can merge it.

@rick-masters Note that this PR moves curl build to sysa, so you don't have to do it in your PR but parts.rst is not yet updated with your extra comments (which will be relevant only when we have fiwix support).

eduardosm: CI passed, but I had forgotten to move the location of curl in `parts.rst`, so pushing again.

stikonas (approved):

PR #274, opened 2023-04-06
Ensure man page is not rebuilt which changes checksum.

### LINKED ISSUE #239: Random failures in sysa help2man 1.36.4
```
help2man-1.36.4: beginning build using script help2man-1.36.4.sh
help2man-1.36.4: getting sources.
help2man-1.36.4.tar.gz: OK
help2man-1.36.4: unpacking source.
help2man-1.36.4: preparing source.
Applying patch: /sysa/help2man-1.36.4/patches/date.patch
patching file help2man.PL
help2man-1.36.4: configuring source.
checking for perl... perl
checking for a BSD-compatible install... /usr/bin/install -c
checking for makeinfo... :
checking for install-info... :
configure: creating ./config.status
config.status: creating Makefile
help2man-1.36.4: compiling source.
perl help2man.PL
Extracting help2man (with variable substitutions)
make help2man
make[1]: Entering directory `/sysa/help2man-1.36.4/build/help2man-1.36.4'
make[1]: `help2man' is up to date.
make[1]: Leaving directory `/sysa/help2man-1.36.4/build/help2man-1.36.4'
./help2man --include=./help2man.h2m \
    --output=help2man.1 ./help2man
make: ./x86/bin/kaem: Shell program not found
make: *** [help2man.1] Error 127
Subprocess error 2
```

rick-masters:
Fixes #239

### DIFF
--- a/sysa/help2man-1.36.4/help2man-1.36.4.sh
+++ b/sysa/help2man-1.36.4/help2man-1.36.4.sh
@@ -17,6 +17,9 @@ src_configure() {
 }
 
 src_compile() {
+    # Ensure man page is not rebuilt which changes checksum
+    touch help2man.1
+
     make MAKEINFO=true
 }
 


fosslinux (changes requested):
But `help2man.1` is pregenerated with `help2man`. It (along with the other `*.1` files in the source) should either be regenerated or removed.

rick-masters:
@fosslinux 
I'm not sure I understand your comment. help2man.1 is already present but it gets regenerated inconsistently. My solution always avoids regenerating it on the rationale that there is no reason to regenerate it and we can keep the existing checksum. Another solution is to remove the file and it will always be regenerated. I'm fine with that but a new checksum would need to be included.

But your comment suggests we either remove it *or* regenerate it. Remove it when? If we remove it before the compile phase it will be regenerated. So that option is to remove it *and* regenerate it, not one *or* the other. Or is there another option I'm missing?

fosslinux:
To some extent, there is little reason to regenerate manpages. However, generally so far, we have regenerated *everything*, manpages included. I'm happy to consider the merits of this - but that's a separate issue to this - I'd prefer to keep this consistent, at least for now.

> Another solution is to remove the file and it will always be regenerated

This is my preference (as long as the checksum will not change, say, next month). One-off checksum changes are basically always OK!

(FWIW: by removing it, I mean removing + patching in a manner that means it does not end up in the final package tarball)

PR #275, opened 2023-04-10
Parallelism

fosslinux:
This should provide a significant speed up to builds on multicore systems.

On my system with --cores 4, there is approx a 2x speedup overall. (Note, this does not scale very well with increased --cores at all, because most things aren't parallelised - but, for example, Guile, GCC 10, Python is much much faster with this).

I haven't tested with QEMU yet.


fosslinux:
Hmm, Guile is a bit slower on few cores. However I also removed `prebuilt/` directory (which used binary blobs...)

stikonas (approved):
Looks good (assuming qemu testing passes).

Though, I wonder if we should introduce (but not in this PR) some make wrapper. E.g. gentoo gas `emake` that calls `make` with some options, if I remember correctly emake=`make ${MAKEOPTS}` where `MAKEOPTS` are user specified options but is usually `-j core_count`. We could also extend it to check for presence of `makeinfo` to pass `MAKEINFO=true` when necessary.

PR #276, opened 2023-04-11
Update dhcpcd to version 9.5.0

eduardosm:
Also fixes broken download link


sysa/dhcpcd-9.5.0/sources
```diff
@@ -0,0 +1 @@
+https://github.com/NetworkConfiguration/dhcpcd/archive/refs/tags/dhcpcd-9.5.0.tar.gz bb8ebda4b250ca8c2faa910205f5ffac66ae83a786b59f5cedbfe7624ad04bc2
```
stikonas, line 1: argh, that is a bit annoying since download hash might change but nothing we can do...

Hopefully we should at least get a 6 months notice (https://github.blog/2023-02-21-update-on-the-future-stability-of-source-code-archives-and-hashes/)

stikonas (approved):

PR #284, opened 2023-04-26
GCC12

fosslinux:
There's not much new here - binutils rebuild is required for a clean GCC 12 bootstrap build - I also do want some of the new features.

libssp stuff is largely inconsequential until we get to adding distributions on top, when it *could* matter.

GCC12 is mostly same as GCC10 - inspection of configure command shows differences.


stikonas (approved):
Might be good to upgrade to GCC 13 and drop patches but let's leave it for follow on PR.

PR #286, opened 2023-05-01
split CI into three (as a temporary solution)

fosslinux:
Fixes the CI length problem.

Draft for testing CI.


fosslinux:
Hmm, the only real motivation for splitting sysc is this CI issue - and this is only a temporary solution - at some point in the near future I'm going to rip out the full build in CI preferably  (need to think a bit more about what to replace it with, but likely only an incremental build).

The split is rather arbitrary. The only consideration I had was CI time.

Anyway, the idea of another sys can be discussed, but separate to this PR.

stikonas (approved):

PR #287, opened 2023-05-01
Fix segfault in yacc

emilytrau:
The patch to heirloom-devtool's yacc used sprintf instead of snprintf. This incompatibility causes a segfault when those code paths are hit. Fixing this allows yacc to compile more grammars, and may allow some packages to drop dependency on bison in favor of yacc.


fosslinux (approved):
LGTM

roconnor:
Seems gawk can build with heirloom-devtools yacc now.

PR #290, opened 2023-05-02
Enable additional tools in coreutils-5.0 phase 1

emilytrau:
These utilities can be enabled with only a small amount of additional patches. In particular sort, expr, uniq,and hostname are used by some ./configure.sh scripts.


emilytrau:
The primary motivation for this is to be able to execute `./configure.sh` scripts at the `tcc-mes` + `bash-2.05b` stage before `musl` is available. While I'm still in the testing phase and don't have a definitive list, `binutil- 2.20.1`'s `./configure.sh` appears to have a dependency on `dirname`, `expr`, `sort`, `uniq` and `hostname`. Not in this list is `factor` and `pr` which were added simply because they didn't throw any compiler errors and didn't take any extra effort. The last two I would be happy to leave out for ~120kb of space saving.

Also worth noting is I discovered the FullSourceBootstrapFromGit project independently implements [a similar patch](https://github.com/schierlm/FullSourceBootstrapFromGit/blob/ca1178a9a02581aac731b10359e639de2a6610c4/part02/coreutils-5.0-dirname-expr-sort.patch) with the same motivations.

```
$ ls -lh tmp/sysa/usr/bin/{dirname,expr,sort,uniq,hostname,factor,pr}
-rwxr-xr-x 1 emilytrau emilytrau  59K May  3 14:08 tmp/sysa/usr/bin/dirname
-rwxr-xr-x 1 emilytrau emilytrau 115K May  3 14:08 tmp/sysa/usr/bin/expr
-rwxr-xr-x 1 emilytrau emilytrau  53K May  3 14:06 tmp/sysa/usr/bin/factor
-rwxr-xr-x 1 emilytrau emilytrau  48K May  3 14:06 tmp/sysa/usr/bin/hostname
-rwxr-xr-x 1 emilytrau emilytrau  71K May  3 14:06 tmp/sysa/usr/bin/pr
-rwxr-xr-x 1 emilytrau emilytrau 113K May  3 14:08 tmp/sysa/usr/bin/sort
-rwxr-xr-x 1 emilytrau emilytrau  74K May  3 14:08 tmp/sysa/usr/bin/uniq
```

fosslinux:
We are a bit more strict about dependency order in live-bootstrap since we require all autotools used in a project before that project's `configure` is used (such as binutils), so I doubt this would actually provide any short-circuits. (Feel free to correct me though!)

@stikonas Are not these binaries overwritten by stage2 anyway? And there's no package for kaem-era programs, so doesn't this change not add any extra space? I may be not thinking straight though

emilytrau:
Thanks for clarifying! That makes sense

emilytrau:
On second thought, given that it wouldn't add extra space to sysa, thoughts on reconsidering this PR? Reason being it would be convenient for me as I wouldn't need to maintain a small patch on the makefile for nixpkgs.

fosslinux (approved):

PR #295, opened 2023-05-10
Kernel bootstrap phase2 v1

### LINKED ISSUE #294: Complete kernel bootstrap with kexec of Linux
The kernel bootstrap ends after building Linux (and rebuilding musl for sysc). 
It needs Fiwix to kexec Linux to complete the process.

A PR is forthcoming.

rick-masters:
Resolves #294


parts.rst
```diff
@@ -11,13 +11,15 @@
 bootstrap-seeds
 ===============
 
-This is where it all begins. We start with the two raw binary seeds ``hex0-seed`` and ``kaem-optional-seed``.
+This is where it all begins.
 
-First, we use those seeds to rebuild themselves.
+A kernel bootstrapping option is used by default at the beginning. The tiny ``builder-hex0-x86-stage1`` binary seed boots, builds and runs the ``builder-hex0-x86-stage2`` kernel. The stage2 kernel has its own built-in shell, the ``hex0`` compiler and the ``src`` tool to load files into its memory file system. In this case the first step is to build the ``hex0-seed`` and ``kaem-optional-seed`` binaries from hex0 source.
```
fosslinux, line 16: can you note that all `builder-hex0` kernels are written in hex0? I know this seems obvious, but just for the sake of absolute clarity.

parts.rst
```diff
@@ -11,13 +11,15 @@
...
 
-Note that all early compilers before ``mes`` are part of `stage0-posix <https://github.com/oriansj/stage0-posix>`_.
+Note that all early shells and compilers before ``mes`` are part of `stage0-posix <https://github.com/oriansj/stage0-posix>`_.
 
-A kernel bootstrapping option is used by default at the beginning. ``hex0-seed`` can be used to compile the ``builder-hex0`` kernel which has its own built-in shell, the ``hex0`` compiler and the ``src`` tool to load files into its file system. ``builder-hex0`` runs stage0-posix and then builds ``mes`` and ``tcc``. It then builds and launches the `Fiwix <https://github.com/mikaku/Fiwix>` kernel which runs the build until Linux takes over.
+``builder-hex0-x86-stage2`` runs as the kernel for building the stage0-posix compilers and then ``mes`` and ``tcc``. Then the `Fiwix <https://github.com/mikaku/Fiwix>` kernel is built and launched. Then Fiwix runs the build until Linux takes over. 
+
+If chroot or bwrap is specified or if a pre-existing kernel is provided for QEMU then we start with the two raw binary seeds ``hex0-seed`` and ``kaem-optional-seed``. We use those seeds to rebuild themselves.
```
fosslinux, line 22: remove "for QEMU" - QEMU is not special in this regard, it could be bare metal or anything else

parts.rst
```diff
@@ -39,9 +41,16 @@ In the first steps we use initial ``hex0`` binary seed to rebuild ``kaem-optiona
 ``hex0`` can be approximated with: ``sed 's/[;#].*$//g' $input_file | xxd -r -p > $output_file``
 
 
-builder-hex0
-============
-By default (when kernel bootstrap is enabled), the ``builder-hex0`` kernel boots from a hard drive and loads an enormous shell script which embeds files (loaded with the ``src`` command) and the initial commands to build ``hex0-seed``, ``kaem-optional-seed``, and the command which launches stage0-posix using ``kaem-optional-seed`` and the stage0-posix launch script ``kaem.x86``. Builder-hex0 is written in hex0 and can be compiled with any one of ``hex0-seed``, ``sed``, the tiny ``builder-hex0-mini`` boot kernel or it can build itself.
+builder-hex0-x86-stage1
+=======================
+
+By default (when kernel bootstrap is enabled), the ``builder-hex0-x86-stage1`` boot loader/compiler boots from a hard drive and loads hex0 source code from disk, compiles, and runs the builder-hex0-x86-stage2 kernel. Builder-hex0-x86-stage1 is written in hex0 and can be compiled with any one of ``hex0-seed``, ``sed``, or the tiny ``builder-hex0-mini`` binary.
```
fosslinux, line 47: very small nit: consistency in terms of code blocks vs plain text, and capitalization

sysa.py
```diff
@@ -193,12 +193,24 @@ def append_srcfs(self, image_file):
...
-            image_file_name)
+        shutil.copyfile(os.path.join('kernel-bootstrap', 'builder-hex0-x86-stage1.bin'), image_file_name)
 
         with open(image_file_name, 'ab') as image_file:
+            current_size = os.stat(image_file_name).st_size
+            while current_size != 510:
+                image_file.write(b'\0')
+                current_size += 1
+            image_file.write(b'\x55')
+            image_file.write(b'\xAA')
```
fosslinux, line 204: What's the purpose of this? (a comment to the effect of whatever the answer is would be great)

fosslinux (commented):
Small things only. I'm very happy with the implementation in general!

sysa.py
stikonas: See also comment in https://github.com/oriansj/bootstrap-seeds/pull/39#issuecomment-1550228515 Maybe it would be good to provide 2 bins? One is 200 byte binary and the other 512 byte binary with MBR signature?

parts.rst
rick-masters: This will be resolved in the next update.

parts.rst
rick-masters: This will be resolved in the next update.

parts.rst
rick-masters: This will be resolved in the next update.

sysa.py
rick-masters: This code will be removed in the next update. It was for padding the boot code to fill a sector and including the MBR signature. The stage1 boot code is now padded in advance. Also, the next update pulls stage1 from stage0-posix.

sysa/kexec-linux/src/kexec-linux.c
```diff
@@ -0,0 +1,91 @@
...
+	}
+	fclose(ramdrive_file);
+
+	/* Perform syscall sync */
+	__asm__ __volatile__(
+			"movl $0x00000024, %%eax\n\t"
+			"int $0x80\n\t"
+			: /* no output */
+			: /* no input */
+	       );
```
stikonas, line 61: Why can't we just call libc function `sync()` from `unistd.h`?

sysa/kexec-linux/src/kexec-linux.c
```diff
@@ -0,0 +1,91 @@
...
+	__asm__ __volatile__(
+			"movl $0x58, %%eax\n\t"
+			"movl $0xfee1dead, %%ebx\n\t"
+			"movl $0x28121969, %%ecx\n\t"
+			"movl $0xcdef0123, %%edx\n\t"
+			"movl $0x00, %%esi\n\t"
+			"int $0x80\n\t"
+			: /* no output */
+			: /* no input */
+	       );
```
stikonas, line 73: Could we use
`syscall(SYS_reboot, int magic, int magic2, int cmd, void *arg)` or maybe even `int reboot(int type)`

sysa/lwext4-1.0.0-lb1/files/make_fiwix_initrd.c
```diff
@@ -267,6 +267,18 @@ int main(int argc, char **argv)
 		printf("ext4_mknod error: %d  \n", err);
 		return EXIT_FAILURE;
 	}
+	printf("ext4_mknod /mp/dev/ram0\n");
```
stikonas, line 270: `puts`?

stikonas (changes requested):
Looks good in general but a few minor comments

sysa/kexec-linux/src/kexec-linux.c
rick-masters: Yes, this can be done but to do this I need to do something I should have done in the first place.
I'll need to compile kexec-linux later in the build with gcc+musl instead of early with tcc+mes/lib.
The reason I compile it early was because it was easy to copy the kexec-fiwix kaem scripts and because, as a small utility program, it didn't fit the helpers.sh/build process used later on for packages.
However, it can be built by manually with a small bash script. It might stick out a bit in run2.sh but I think its more correct than what I'm doing now because it will be built closer to where it is used and will allow using the library calls you have suggested.
I have this working and my next update will include this change.

stikonas: You can just put `gcc` command into `kexec-linux.sh` script in `src_compile()` function and I think it will just work. And then you also need `src_install()` step that copies it to "${DESTDIR}${PREFIX}/bin/".

rick-masters: I can try to use the build command but I'll need to build a package with a version number because there is no way to opt out of that. I'm not sure its worth it as the package is unlikely to be used outside live-bootstrap. But I'll start working on it if you think it makes sense.

stikonas: Oh yes, that's a limitation of `helpers.sh`. Perhaps you can just use dummy version, e.g. `kexec-linux-1`

fosslinux: Version also makes sense because of `--repo`. The version doesn't need to be significant, it can just be incremented when there's a change.

rick-masters: This should be resolved. kexec-linux is a package in the recent update.

sysa/lwext4-1.0.0-lb1/files/make_fiwix_initrd.c
rick-masters: puts is used where possible in the recent update.

sysa/kexec-linux/src/kexec-linux.c
rick-masters: sync() is now used.

sysa/kexec-linux-1.0.0/files/kexec-linux.c
```diff
@@ -28,15 +29,15 @@ int main(int argc, char **argv) {
 
 	/* Write length of kernel */
 	if (stat(kernel_file_name, &stats) == 0) {
-		fwrite(&stats.st_size, sizeof(stats.st_size), 1, ramdrive_file);
+		fwrite(&stats.st_size, 4, 1, ramdrive_file);
```
stikonas, line 32: is this change `sizeof() -> 4` intentional?

stikonas (approved):

sysa/kexec-linux-1.0.0/files/kexec-linux.c
rick-masters: Yes, this must be a 32 bit integer because that is what Fiwix reads from the ram drive. This code does make a little-endian assumption for larger integers but it works. I probably should have cast stats.st_size to a uint32_t variable before writing. I can make that change ... but it might take a couple hours to test and generate new checksum.

stikonas: I guess let's keep it like that then.

rick-masters: Well I've coded an improvement using a cast to uint32_t. It is more portable although big-endian usage seems pretty unlikely. I'm willing to submit it later after testing if this PR is still open.

stikonas: Sure feel free to submit it, we haven't merge it yet, though I guess we are close to being done here. I'm also running some testing now...

stikonas (approved):

stikonas:
Anyway, looks good, let's merge it and any further improvements can be added with the follow-on PRs.

PR #299, opened 2023-06-01
gzip: Regenerate crc table in util.c using makecrc

### LINKED ISSUE #186: Regen CRC tables in gzips
`makecrc.c` included in the gzip tarball can be used to regenerate the hard-coded CRC tables.

pder:
Closes #186


stikonas (approved):
Looks fine in general, but maybe it makes sense to build gnu patch before gzip then? @fosslinux  what do you think?

fosslinux:
Yeah, patch before gzip makes sense to me, as long as `untar` and `ungz` are robust enough to deal with the patch tarball. (I'll test this and report back).

fosslinux:
Yes - it works. I'll push that change to master then this PR can be rebased on top of that.

pder:
This branch has been updated so that we build make, patch, and gzip in that order.  Patch is extracted using ungz and untar.  I use patch instead of simple-patch for updating gzip code,  I added a separate commit for cleaning up the simple-patch hard coded paths containing architecture.

fosslinux (approved):
Thanks!

PR #301, opened 2023-06-12
Update to musl 1.2.4.

stikonas:
musl 1.2.4 introduces minor breakage (for now e2fsprogs and autogen are affected) that for now can be workarounded with `CPPFLAGS=-D_LARGEFILE64_SOURCE=1` though this might be removed in future versions of musl. I've decided to stop building e2defrag as we don't really need it.


fosslinux (approved):
I'll just let CI finish, but LGTM.

PR #305, opened 2023-06-15
Boot Fiwix and Linux without serial console on bare metal

pder:


stikonas (approved):
That seems to only change Fiwix kernel options. Or does it also propagate to Linux?

fosslinux (commented):
Similar change to go_sysb would be good (so that console=ttyS0 is not used for sysc).

pder:
> That seems to only change Fiwix kernel options. Or does it also propagate to Linux?

kexec_cmdline are the parameters passed to Linux.

kexec_cmdline=\"init=/init console=ttyS0\"

pder:
> Similar change to go_sysb would be good (so that console=ttyS0 is not used for sysc).

OK, so would this be for the case we are using the bare-metal option and providing a kernel?

stikonas:
No, just the different transition. You updated builder-hex0 to fiwix but there is Fiwix to Linux here: https://github.com/fosslinux/live-bootstrap/blob/master/sysa/run2.sh#L38

pder:
> No, just the different transition. You updated builder-hex0 to fiwix but there is Fiwix to Linux here: https://github.com/fosslinux/live-bootstrap/blob/master/sysa/run2.sh#L38

The Fiwix to Linux transition happens here: 
https://github.com/fosslinux/live-bootstrap/blob/master/sysa/run2.sh#L29

rickmasters created kexec-linux that is what sets up Linux for kexecing from Fiwix

stikonas:
> > No, just the different transition. You updated builder-hex0 to fiwix but there is Fiwix to Linux here: https://github.com/fosslinux/live-bootstrap/blob/master/sysa/run2.sh#L38
> 
> The Fiwix to Linux transition happens here: https://github.com/fosslinux/live-bootstrap/blob/master/sysa/run2.sh#L29
> 
> rickmasters created kexec-linux that is what sets up Linux for kexecing from Fiwix

Oh indeed, we have a different path for bootstrapping. That other kexec only happens if we start with another linux kernel.

Though I'm still trying to understand where do we pass linux command line to kexec-linux..

fosslinux:
Huh, but where *is* the cmdline for sysc in kernel bootstrap mode? Does it exist?

stikonas:
> Huh, but where _is_ the cmdline for sysc in kernel bootstrap mode? Does it exist?

Yes, see irc logs on #bootsrappable. On Fiwix future kexec cmdline has to be set in advance inside Fiwix cmdline (kexec_cmdline parameter)

stikonas:
https://github.com/mikaku/Fiwix/issues/29 documents this a bit.

fosslinux (changes requested):
Consensus is could you please add the change to `go_sysb` in `sysa/run2.sh` as well? For when kernel bootstrap is not used.

fosslinux:
You would use GRUB or something similar and then `linux path/to/kernel` and `initrd path/to/initramfs`.

stikonas:
Hmm, it's probably fine to just test this in qemu... We never really tested GRUB before anyway. Though if you can, that would be an interesting test.

fosslinux:
I have once before, but it's been a fair while since I've tested it..

stikonas (approved):

PR #311, opened 2023-07-15
Mount a few missing device nodes in chroot bootstrap modes

doraskayo:
## Context

This closes a gap between the chroot and non-chroot bootstrap modes, which was originally created in 7ecad37 because BuildStream's sandbox was missing some of these devices.

With the latest version of buildbox-run-bubblewrap, the missing devices were made available in BuildStream's sandbox and there is no longer a need to keep this gap. In fact, compatibilty with newer BuildStream versions now requires these devices to be present.

This change also partially reverts d0a5221, which had to remove /dev/ptmx as a consequence of the aforementioned gap.

## Testing

Tested and confirmed that package hashes are stable with the bwrap, chroot and QEMU kernel bootstrap modes. Also with BuildStream through the freedesktop-sdk-binary-seed project.


doraskayo:
Marked as draft until CI succeeds for freedesktop-sdk-binary-seed: https://gitlab.com/freedesktop-sdk/freedesktop-sdk-binary-seed/-/merge_requests/60

rootfs.py
```diff
@@ -214,6 +217,9 @@ def bootstrap(args, system_a, system_c, tmpdir):
                          '--dev-bind', '/dev/zero', '/dev/zero',
                          '--dev-bind', '/dev/random', '/dev/random',
                          '--dev-bind', '/dev/urandom', '/dev/urandom',
+                         '--dev-bind', '/dev/ptmx', '/dev/ptmx',
```
nanonyme, line 220: How about we make this as well use --dev?

doraskayo: While I'm not entirely against doing this, I'm not very eager to do it either.

It would require adding bwrap's logic behind `--dev` in `populate_device_nodes()` to reduce the risk of getting different package hashes in different bootstrap modes. Basically, creating even more more device nodes and symlinks.

However, the main issue is that bwrap may freely change the device nodes it creates when `--dev` is used, potentially resulting in getting different package hashes in different bwrap versions.

I personally think buildbox-run-bubblewrap made a mistake by choosing to use `--dev` to add its missing device nodes, and I'm not sure whether it's a good idea to do the same.

nanonyme: Well, this is still missing multiple important files like /dev/fd, /dev/fd/0 etc.

fosslinux: A handful of `./configure` scripts check for the presence of `/dev` devices to see whether the system supports something. If bwrap just arbitrarily adds devices that would lead to checksum changes.

/dev/fd etc aren't currently used in live bootstrap, we mknod everything we need.

nanonyme: @fosslinux the thing is, this now differs from BuildStream as it has more items in /dev (due to using --dev). While currently both result on same hashes, this may not be guaranteed if new packages get added to live-bootstrap.

nanonyme: Those extra devices are definitely needed for post-live-bootstrap building of various packages.

nanonyme: Anyway, feel free to resolve this and regards this as bikeshedding. I can't apparently resolve this discussion.

doraskayo: For now, I prefer to have full control over the devices we expose in live-bootstrap. This can change if bubblewrap provides us certainty/stability regarding which devices are created in the sandbox when the `--dev` option is used.

I feel that the same line of thought should have lead the thought process around buildbox-run-bubblewrap, but unfortunately it hasn't. The fallout from that decision should be a BuildStream issue, not a live-bootstrap issue.

doraskayo:
freedesktop-sdk-binary-seed CI is green: https://gitlab.com/freedesktop-sdk/freedesktop-sdk-binary-seed/-/pipelines/932520692

This should now be ready.

stikonas (approved):

PR #314, opened 2023-07-25
gcc 13.1.0: don't enable default ssp or pie

pder:
The combination of --enable-default-ssp and the modifications to musl were causing gcc to produce binaries that segfault on startup.  Also turn off --enable-default-pie.  These features can be enabled later when a future toolchain is built.


fosslinux:
Was it confirmed that PIE was a cause of segfaults?

pder:
It has not been confirmed yet that pie was the reason for the segfaults doras experienced.

stikonas (approved):

fosslinux:
I'm not convinced this is the end of the story, but I need to do a bit more digging.

doraskayo:
@fosslinux, @pder, @stikonas: I'm happy to share that this PR solved the segmentation faults we've seen with GCC 13.1.0. See: https://gitlab.com/freedesktop-sdk/freedesktop-sdk-binary-seed/-/pipelines/951421539

I've made two tests, one only with the removal of `--enable-default-pie`  and another with the removal of both `--enable-default-pie`  and `--enable-default-ssp`. While the first reduced the number of segmentation faults seen, the second was required to solve the issue entirely.

PR #315, opened 2023-08-01
Add dev nodes to sysc

ajherchenroder:
This change adds three SATA disks with three partitions each along with three NVME drives with three partitions and a optical drive to sysc.
Fixes Issue# 312

### DIFF
--- a/sysc/after.sh
+++ b/sysc/after.sh
@@ -4,7 +4,35 @@
 # SPDX-License-Identifier: MIT
 
 # Replace this hook if you wish to do more
-
+# Add Dev Nodes 
+# Disk Nodes
+# 3 disks 3 partitions
+major=0
+alpha="a b c" 
+# For each disk...
+for a in ${alpha}; do
+        mknod -m 600 "/dev/sd${a}" b 8 "$((major*16))"
+        #For each partition do...
+	minor=1 
+        for p in $(seq 3); do
+            mknod -m 600 "/dev/sd${a}${p}" b 8 "$((major*16+minor++))"
+        done
+	((major++))
+done
+# NVME Nodes
+# 3 NVME disk with 3 partitions
+major=0 
+# For each disk...
+for a in $(seq 3); do
+	mknod -m 600 "/dev/nvme${a}" c 259 0 # NVME CHAR node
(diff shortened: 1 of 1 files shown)


sysc/after.sh
```diff
@@ -4,6 +4,28 @@
...
+        mknod -m 600 "/dev/sd${a}" b 8 "$((minor++))"
+        # For each partition...
+        for p in $(seq 3); do
+            mknod -m 600 "/dev/sd${a}${p}" b 8 "$((minor++))"
+        done
+    done
+# NVME Nodes
+minor=0 # 3 NVME disk with 3 partitions
+    # For each disk...
+    for a in ${seq 3}; do
```
stikonas: Should it be `$(seq 3)`?

stikonas (changes requested):
Could you please squash it into a singled commit?

ajherchenroder:
> Could you please squash it into a singled commit?
The last thing I worked on in GitHub squashed the commits automatically. Is there a manual way to do it?

fosslinux:
You can do an interactive rebase at the terminal (`git rebase -i HEAD~3`, for example to work with the last 3 commits), and then the comment text will explain it.

stikonas (approved):
I guess looks good. Though I'm a bit confused by indentation here.

@fosslinux what do you think?

fosslinux:
agreed, I do understand the indenting to indicate sections of code however generally i'd prefer line breaks rather than idents.

ajherchenroder:
I don’t mind flattening the spacing but but I was attempting to mirror the formatting in run.sh in sysb. Version 0 of this change was just dropping parts of that code in. When I went back and modified it, I kept the formatting to try to keep the look and feel. Let me know what you want me to do.

fosslinux:
Let's flatten the spacing (but keeping usual indentation for loops, ifs etc).

We follow generally "logic indenting", where each control block indents. That means that when you copy code from one file to another the indentation probably needs to be changed. (Excuse me if you were already aware of this but I think something isn't perfectly clear here.)

nanonyme:
Just a nit but there's a typo in description which prevents this from actually linking to and closing the related issue.

fosslinux (approved):
LGTM

stikonas (approved):

Issue #321, opened 2023-10-14
Python 2.5 build failures with parallel build on


nanonyme:
Noticed at https://gitlab.com/freedesktop-sdk/freedesktop-sdk-binary-seed/-/jobs/5292964461 with max jobs 12. Unclear how well reproducible.

nanonyme:
@doraskayo suggested might be related to https://bugs.python.org/issue10013

nanonyme:
https://github.com/python/cpython/commit/9828b56347f6171a532dce28c9dc5ab9bd2c8139.patch might be the real fix. The patch is under PSF.

nanonyme:
It looks like fix was backported so it does not impact our 3.1.5.

nanonyme:
Based on brief glance same bug may be present in 2.3.5. 2.0.1 tag does not exist in CPython git repository so didn't check yet.

nanonyme:
@fosslinux how does copyright tag work with these? It looks like we would end up mostly using upstream patches.

nanonyme:
Okay. Given the patch should be fine with just NEWS change *removed*, maybe it's not meaningful to add me. Real change is property of said person.

doraskayo:
> Based on brief glance same bug may be present in 2.3.5. 2.0.1 tag does not exist in CPython git repository so didn't check yet.

Wait, what about these other versions? Do we need a separate issue for those?

nanonyme:
I dunno, I have never seen this problem on anything else than 2.5.6.

PR #325, opened 2023-11-06
Fix Python 2.5.6 for parallel pgen invocation

nanonyme:
Backported from https://github.com/python/cpython/commit/9828b56347f6171a532dce28c9dc5ab9bd2c8139.patch
Fixes https://github.com/fosslinux/live-bootstrap/issues/321


nanonyme:
@fosslinux looks like build is failing as part of pkg-config removal. I'm running a pipelien on our side at https://gitlab.com/freedesktop-sdk/freedesktop-sdk-binary-seed/-/jobs/5467602634

nanonyme:
We are mirroring our sources so the removal doesn't really impact us. We are using method for obtaining sources that is similar to bwrap method.

nanonyme:
Unfortunately it looks like this patch changes hashes. I will need to find out if they are changed in a reproducible way.

fosslinux:
Yes, the pkg-config thing is expected to fail builds at the moment.

Many things can change hashes, I'm not particularly concerned that hashes are changed - just that the new hash is always the same :)

nanonyme:
Does the pkg-config situation effectively mean nothing can be merged for the time being?

fosslinux:
The CI should still pass, because it's using download-distfiles.sh (external distfiles)...

Things can still be merged.

nanonyme:
This CI ironically would most likely be considerably faster with parallel build on.

nanonyme:
@fosslinux it seems this new hash reproduced multiple times so I'm hopeful that it persists. I suspect there is one extra file (the timestamp file) as a result of the patch.

fosslinux:
The CI already runs with `--cores 2`, which is what the runners have available.. https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources

fosslinux (approved):
LGTM

PR #328, opened 2023-11-13
[WIP] Make it possible to reboot into the final bootstrapped system

Googulator:
GRUB build still needs to be converted into a true build step, with hash checking and proper package management.

### DIFF
--- a/rootfs.py
+++ b/rootfs.py
@@ -256,7 +256,7 @@ print(shutil.which('chroot'))
             system_a.prepare(create_initramfs=True, kernel_bootstrap=True)
             run(args.qemu_cmd,
                 '-enable-kvm',
-                '-m', "4G",
+                '-m', "8G",
                 '-smp', str(args.cores),
                 '-no-reboot',
                 '-drive', 'file=' + os.path.join(system_a.tmp_dir, 'sysa.img') + ',format=raw',
--- a/sysa.py
+++ b/sysa.py
@@ -222,4 +222,4 @@ class SysA(SysGeneral):
 
         # extend file up to desired size
         with open(image_file_name, 'ab') as image_file:
-            image_file.truncate(16384 * megabyte)
+            image_file.truncate(32768 * megabyte)
--- a/sysa.py
+++ b/sysa.py
@@ -220,8 +220,6 @@ class SysA(SysGeneral):
             image_file.write(b'\0' * round_up)
         current_size += round_up
 
-        # fill file with zeros up to desired size, one megabyte at a time
+        # extend file up to desired size
         with open(image_file_name, 'ab') as image_file:
-            while current_size < 16384 * megabyte:
-                image_file.write(b'\0' * megabyte)
-                current_size += megabyte
+            image_file.truncate(16384 * megabyte)
--- a/sysa/fiwix-1.4.0-lb3/sources
+++ b/sysa/fiwix-1.4.0-lb3/sources
@@ -1,1 +1,1 @@
-https://github.com/rick-masters/Fiwix/releases/download/v1.4.0-lb3/fiwix-1.4.0-lb3.tar.gz 525a24e32571e574da76e2a8c898f4eab371dfb85302d545c498a21c8028283d
+https://github.com/Googulator/Fiwix/releases/download/v1.4.0-lb3-wip/fiwix-1.4.0-lb3.tar.gz ee8b811b8280096ed143c9c5b2f6179a029b194de2cb89a1464895fba581c8e5
--- a/sysa/kexec-fiwix/src/kexec-fiwix.c
+++ b/sysa/kexec-fiwix/src/kexec-fiwix.c
@@ -123,7 +123,7 @@ int main() {
 
 	/* Set memory map info */
 	pmultiboot_info->mmap_addr = next_avail_mem;
-	pmultiboot_info->mmap_length = 7 * sizeof(multiboot_memory_map_t);
+	pmultiboot_info->mmap_length = 8 * sizeof(multiboot_memory_map_t);
 	multiboot_memory_map_t *pmultiboot_memory_map = (multiboot_memory_map_t *) next_avail_mem;
 
 	pmultiboot_memory_map->size = sizeof(multiboot_memory_map_t) - sizeof(multiboot_uint32_t);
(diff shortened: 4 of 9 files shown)


fosslinux:
I have a few questions with this PR.

1. Why is there: increased RAM requirement to 8G, increased kexec target size to 256MB, 8G swap in sysc? This is not something that is particularly desirable arbitrarily.
2. What is "PAE passthrough support"?
3. Why does the entirety of sysa need to be copied into sysc?

(it is still a WIP is a valid answer to all of these!)

I like the idea of making the final image bootable, as this would allow one to come back to their live-bootstrap environment once it finishes.

If you don't mind, I might cherry-pick a few of these earlier fixes, namely [Speed up sysa.img generation](https://github.com/fosslinux/live-bootstrap/pull/328/commits/d67a3d2bb4d0c2526627e5e5ba3b2368a5202008) [Change openssl source URL because the old one redirects to HTTPS](https://github.com/fosslinux/live-bootstrap/pull/328/commits/2fe14bbcb1a5b0acf469fce5c48eb18fcf269609) [Start Linux rootfs partition at 1M, not sector 1](https://github.com/fosslinux/live-bootstrap/pull/328/commits/5bd1bad62e0c36b83a96daea15baffdd9c34818c) separately to this PR, as they aren't really related to this PR but are still useful fixes.

I have a fairly large PR coming in the next few days that unfortunately will conflict a fair bit with this PR but should also allow this PR to be a bit cleaner.. apologies in advance, I will help refactor once that is in.

Googulator:
While we don't technically need all of sysa to be brought into sysc, to make the final image bootable, it's necessary to bring over at least the Linux kernel. Bringing over the rest was something I did mainly so I can inspect the binaries created during the earlier stages, i.e. I wanted a "nothing gets left behind" build. Of course, all of this should be eventually put behind a flag, but for this draft, I haven't bothered to do that yet. (In fact, I'm planning to securely erase srcfs from both RAM and disk once it's no longer needed, as part of an additional anti-Trusting-Trust measure.)

The additional RAM and swap are only tangentially related to getting a bootable image - I was testing on a 16-thread CPU, and I had several bootstrap attempts fail due to OOM in sysc. Eventually, I want to make both swap (whether or not to create it, and if so, how much) and RAM configurable, as well as the size of the image created. (RAM is already configurable when kernel bootstrap is disabled, but hardcoded when it's enabled.)

"PAE passthrough" is really a fix for what I consider to be a bug in the Fiwix kernel. Fiwix doesn't support PAE (not that it would have much use for it with its 1GB RAM hard limit), but Linux does, and PAE is a requirement for the 32-bit Linux kernel to manage more than 4GB of RAM. PAE is already enabled in the kernel config, but Fiwix's memory map parsing code assumes that nothing downstream will support PAE, and deletes memory blocks >4GiB from the memory map. The PAE passthrough patches fix this: Fiwix now keeps track of memory blocks in PAE address space, internally ignoring them, but passing them on to any kexec target that can use them. (The code for this is [here](https://github.com/rick-masters/Fiwix/compare/fiwix-1.4.0-lb3...Googulator:Fiwix:fiwix-1.4.0-lb3).)

fosslinux:
> While we don't technically need all of sysa to be brought into sysc, to make the final image bootable, it's necessary to bring over at least the Linux kernel. Bringing over the rest was something I did mainly so I can inspect the binaries created during the earlier stages, i.e. I wanted a "nothing gets left behind" build

OK, that makes sense.

> In fact, I'm planning to securely erase srcfs from both RAM and disk once it's no longer needed, as part of an additional anti-Trusting-Trust measure

Could you elaborate on this? I'm not certain how this helps with trusting trust (I feel I may be misinterpreting this)

> "PAE passthrough" is really a fix for what I consider to be a bug in the Fiwix kernel. Fiwix doesn't support PAE (not that it would have much use for it with its 1GB RAM hard limit), but Linux does, and PAE is a requirement for the 32-bit Linux kernel to manage more than 4GB of RAM. PAE is already enabled in the kernel config, but Fiwix's memory map parsing code assumes that nothing downstream will support PAE, and deletes memory blocks >4GiB from the memory map. The PAE passthrough patches fix this: Fiwix now keeps track of memory blocks in PAE address space, internally ignoring them, but passing them on to any kexec target that can use them. 

Ahh, alright, understood. That's actually very useful, thank you!

Googulator:
The anti-Trusting-Trust plan involves modifying builder-hex0 stage 1 (the boot sector) to print out the hex0 code of stage 2 as it gets loaded, and then having stage 2 likewise print out the contents of files in the srcfs (excluding gz/bz2 archives - those can get printed instead as they are decompressed). This way, everything in srcfs and before gets printed out (& can be securely recorded on a separate trusted medium where it can be read back for auditing) before it has a chance to execute and therefore go back to modify its own source; with the exception of the stage 1 boot sector (limited to 512 bytes). Then, as we jump into sysc, we erase srcfs, before we start downloading any other source code.

At the end of sysc, we redownload the sources needed for sysa, and build another sysa.img for bare metal usage - if the initial bootstrap was bare metal, this should be bit-for-bit identical to the one we originally built on our host system, and since we erased the original srcfs, it must have been really built from source, and not merely stashed away & then revealed by (potentially malicious) new code downloaded within sysc.

Of course, since we can't quite trust any tool either on the original host system, or in the bootstrapped environment, with comparing the 2 sysa images, we instead compare by writing the 2nd sysa.img to another drive, and bootstrapping again from it (ideally on new hardware never used for bootstrapping before, to exclude the possibility of a backdoor from a prior compromised bootstrap getting stashed within BIOS or other hidden storage). Record the output securely as before. This 2nd bootstrap can then be stopped once we can verify that sysc has been reached.

If we now compare the records from both bootstraps, and the sources printed out match, revealing no backdoor, then we can be certain the sysa.img we built is truly clean - even if our original host system was compromised.

Googulator:
A heads up: https://github.com/fosslinux/live-bootstrap/pull/328/commits/5bd1bad62e0c36b83a96daea15baffdd9c34818c works fine in qemu, but I've seen it fail on bare metal with "I don't like these options". It seems when the medium is an actual ATA hard drive that can be queried for geometry, sfdisk will reject -S32 -H64 unless --force is also specified.

Googulator:
All of this was merged in other PRs in an updated form.

PR #331, opened 2023-11-14
Mes 0.25

stikonas:
This updates mes to version 0.25 but does not yet touch bootstrappable tcc tarball which should also be updated (will be done in follow on PR later).


stikonas:
There are a couple of unused variables in tcc build script (e.g. HAVE_LONG_LONG), that will be used later when I push more of my changes to tcc build.

PR #333, opened 2023-11-19
Update to newer bootstrappable tcc snapshot.

stikonas:
Also reworked some tcc stuff. Building it 6 times is overkill. First tcc-mes version already supports all needed features and 3 more rebuilds are sufficient to reach a stable point (in terms of output hash).


fosslinux (approved):
LGTM, I fear it is likely that checksums for other arches may rot, but time will tell...

stikonas:
> LGTM, I fear it is likely that checksums for other arches may rot, but time will tell...

Possibly, but maybe we can at least update them before release. (Also as it is this would only run on real HW on other arches, now with qemu as it is missing `--ro-bind-try /path/to/qemu /path/to/qemu`)

PR #336, opened 2023-11-27
Return process from run_as_root.

### LINKED ISSUE #335: chroot broken by new run_as_root helper
When the --chroot option is used rootfs.py attempts get the path to the chroot binary by running a small python3 script.
But the new run_as_root helper does not return the process object and so attempting to access stdout of the python3 process fails.

```
$ ./rootfs.py --chroot
Bootstrapping x86 -- SysA
subprocess.run(['sudo', 'python3', '-c', "\nimport shutil\nprint(shutil.which('chroot'))\n"], check=True, {'capture_output': True}
CompletedProcess(args=['sudo', 'python3', '-c', "\nimport shutil\nprint(shutil.which('chroot'))\n"], returncode=0, stdout=b'/usr/sbin/chroot\n', stderr=b'')
Traceback (most recent call last):
  File "./rootfs.py", line 268, in <module>
    main()
  File "./rootfs.py", line 161, in main
    bootstrap(args, system_a, system_c, tmpdir)
  File "./rootfs.py", line 171, in bootstrap
    chroot_binary = run_as_root('python3', '-c', find_chroot,
AttributeError: 'NoneType' object has no attribute 'stdout'
```

rick-masters:
Fixes #335

### DIFF
--- a/lib/utils.py
+++ b/lib/utils.py
@@ -28,9 +28,9 @@ def run(*args, **kwargs):
 def run_as_root(*args, **kwargs):
     """A helper for run that invokes sudo when unprivileged"""
     if os.geteuid() != 0:
-        run("sudo", *args, **kwargs)
+        return run("sudo", *args, **kwargs)
     else:
-        run(*args, **kwargs)
+        return run(*args, **kwargs)
 
 def create_disk(image, disk_type, fs_type, size):
     """Create a disk image, with a filesystem on it"""
--- a/lib/utils.py
+++ b/lib/utils.py
@@ -29,8 +29,7 @@ def run_as_root(*args, **kwargs):
     """A helper for run that invokes sudo when unprivileged"""
     if os.geteuid() != 0:
         return run("sudo", *args, **kwargs)
-    else:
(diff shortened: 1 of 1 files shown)


nanonyme:
Oops, good catch. I wonder why CI didn't fail....

lib/utils.py
```diff
@@ -28,9 +28,9 @@ def run(*args, **kwargs):
 def run_as_root(*args, **kwargs):
     """A helper for run that invokes sudo when unprivileged"""
     if os.geteuid() != 0:
-        run("sudo", *args, **kwargs)
+        return run("sudo", *args, **kwargs)
     else:
```
nanonyme, line 32: Seems linter wants you to remove the else as redundant now that there's an early return.

rick-masters: Thanks. Fixed.

rick-masters:
> Oops, good catch. I wonder why CI didn't fail....

Unfortunately CI does not test with chroot option.

fosslinux (approved):

PR #339, opened 2023-11-28
Add small bubblewrap like program as a build step

MaxHearnden:
This allow the user to perform a rootless build without bubblewrap on machines with user namespaces enabled

### DIFF
--- a/rootfs.py
+++ b/rootfs.py
@@ -31,8 +31,9 @@ def create_configuration_file(args):
     config_path = os.path.join('sysa', 'bootstrap.cfg')
     with open(config_path, "w", encoding="utf_8") as config:
         config.write(f"FORCE_TIMESTAMPS={args.force_timestamps}\n")
-        config.write(f"CHROOT={args.chroot or args.bwrap}\n")
+        config.write(f"CHROOT={args.chroot or args.wrap or args.bwrap}\n")
         config.write(f"CHROOT_ONLY_SYSA={args.bwrap}\n")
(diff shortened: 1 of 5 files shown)


fosslinux:
I understand the idea/implementation in principle. Would you mind outlining any particular, specific usecases for this?

MaxHearnden:
It's to reduce the binary seed required for an unprivileged build to just hex0+kaem-optional and can be used with the following commands
```
cp sysa/stage0-posix/src/* .
cp sysa/after_wrap.kaem after.kaem

# This stage could be integrated into after_wrap.kaem
echo "CHROOT=True
CHROOT_WRAP=True" >> sysa/bootstrap.cfg

./bootstrap-seeds/POSIX/x86/kaem-optional-seed
```

oriansj:
I like it but 2 minor details:
1) the syscalls should added to M2libc such that the code becomes more standard C.
2) the wrap.c program belongs in mescc-tools-extra which live-bootstrap does already use and will allow other bootstrapping projects to leverage your work.

great work

fosslinux:
If this could be

1. rebased on master
2. updated to reflect that wrap.c is in mescc-tools-extra

I think I'm happy to include this in principle.

Googulator:
#436 will bring in wrap.c into live-bootstrap, although currently unused

MaxHearnden:
Replaced by https://github.com/fosslinux/live-bootstrap/pull/479

PR #340, opened 2023-11-28
Kaem-era improvements

fosslinux:
- Use uppercase variables as we do in bash-era. The use of lowercase variables and uppercase variables in kaem and bash respectively didn't really make very much sense.
- *Always* use `/usr/lib` and `/usr/include` for musl, and `/usr/lib/mes` and `/usr/include/mes` for mes libc.
  - Before it was a bit haphazard and there was no clear rules, things just kinda moved around as development occurred. Let's standardize this.

I'm not sure if this actually works. Hypothetically it should but haven't tested.

Cherry-picked from #334


fosslinux:
I would like not to regenerate checksums now and just do it in one big updating checksum commit in #334.

stikonas (approved):

Issue #323, opened 2023-10-22
simple-patch issues


stikonas:
* `simple-patch` needs to have its checksum computed.
* [DONE] install it to `/usr/bin` rather than `/x86/bin/`
* [DONE] Pass  `--architecture ${ARCH}` rather than x86 to `M2-Mesoplanet` when building it.

stikonas:
This is now partially as part of riscv64 addition. The only missing thing is checksum.

PR #341, opened 2023-11-28
Early tools in subdirectories

fosslinux:
Make packages out of `checksum-transcriber`, `simple-patch`. also give kexec-fiwix a version number and rename `sysa.ext2` to `fiwix.ext2`

Prep for & cherry-picked from #334 

Depends on #340


sysa/simple-patch-1.0/simple-patch-1.0.checksums
```diff
@@ -0,0 +1 @@
+bf49a897e96fe4052904aaedc8281f6849e3dc1836899d393c43726cc8f5e92e  /usr/bin/simple-patch
```
stikonas, line 1: I believe this closes #323 . Maybe worth adding it in commit message to auto-close it.

sysa/simple-patch-1.0/simple-patch-1.0.kaem
```diff
@@ -0,0 +1,20 @@
...
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+set -ex
+
+# Build & install
+M2-Mesoplanet --architecture ${ARCH} -f src/simple-patch.c -o ${BINDIR}/simple-patch
+
+# Checksums
+if match x${UPDATE_CHECKSUMS} xTrue; then
+    sha256sum -o ${pkg}.checksums \
```
stikonas, line 14: Maybe let's use arch-specific checksum file here? I know riscv64 is not fully supported, but it can reach this point.

stikonas (approved):
Mostly good. I left a couple of comments that would be good to fix.

(Feel free not to add riscv64 checksum file though... I can do it at some point later)

PR #343, opened 2023-12-05
Build stat for coreutils

fosslinux:
dep of #334 

no checksum update for big checksum update in #334


stikonas (approved):
I guess looks good. But what is the use of `stat`? Is it needed by some `configure` script?

I was actually thinking of moving the other direction and removing all the unnecessary stuff that we have here...

stikonas:
> But what is the use of `stat`?
Oh I guess it's for initrd sizes?

fosslinux:
I added it for [easily determining major and minor numbers of device nodes](https://github.com/fosslinux/live-bootstrap/blob/simplify/steps/jump/linux.sh#L19).

PR #354, opened 2023-12-17
Python script fixes and improvements

Googulator:
- Support specifying size of the virtual disk for qemu
- For bare metal, only create a minimal sized boot disk image
- Don't create external.img, unless it's actually needed
- Use truncate() for extending disk images to the desired size (faster than writing zeros)
- Make source_manifest.py work again
- Make -qr compatible with kernel bootstrap (will need fix in kexec-fiwix to actually use more memory)
- Use correct partition alignment (only in the Python scripts yet, move_disk.sh needs to be also fixed)
- Reserve the first 1GB on boot disks (again, needs to be done in move_disk.sh too)
- Fix bootstrap.cfg creation/cleanup (broken by residual sysa reference)
- Avoid including a truncated copy of the srcfs image in itself
- Print the correct paths to the images in bare metal mode
- Rework source manifest generation to be based on the bootstrap manifest
- Support predownloading only the sources that are needed before network
- Eliminate steps/pre-network-sources, decide based on the bootstrap manifest instead


fosslinux (commented):
Could you please also;

- fix pylint errors
- squash any "Fix" commits that fix an error introduced in an earlier commit (within this PR) into that commit

lib/generator.py
```diff
@@ -313,9 +313,16 @@ def get_source_manifest(cls):
 
         # Find all source files
         steps_dir = os.path.join(cls.git_dir, 'steps')
-        for file in os.listdir(steps_dir):
-            if os.path.isdir(os.path.join(steps_dir, file)):
-                sourcef = os.path.join(steps_dir, file, "sources")
+        with open(os.path.join(steps_dir, 'manifest'), 'r', encoding="utf_8") as file:
+            for line in file:
+                if pre_network and line.strip().startswith("improve: ") and "network" in line:
```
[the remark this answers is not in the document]
Googulator: Will do, together with a similar comment for the significance of "build: bash*".

Googulator:
Is there a better way to do that in Git than redoing the entire series by hash, and force-pushing? I.e. something similar to Mercurial's histedit.

rootfs.py
```diff
@@ -200,7 +210,8 @@ def bootstrap(args, generator, tmpdir):
                          init)
 
         if not args.internal_ci or args.internal_ci == "pass2" or args.internal_ci == "pass3":
-            shutil.copy2(os.path.join('sysa', 'bootstrap.cfg'),
+            # TODO sysa/sysc no longer exists
+            shutil.copy2(os.path.join('steps', 'bootstrap.cfg'),
                          os.path.join('tmp', 'sysa', 'sysc_image', 'usr', 'src', 'bootstrap.cfg'))
             run('bwrap', '--unshare-user',
```
stikonas, line 216: It might also be nice to somehow allow passing extra bwrap arguments. I.e. passing qemu binary with `--ro-bind-try` allows testing bootstrap on other arches.

fosslinux:
> Is there a better way to do that in Git than redoing the entire series by hand, and force-pushing? I.e. something similar to Mercurial's histedit.

`git rebase -i HEAD~n`, where `n` is the number of commits back you'd like to adjust the history. Happy to walk you through this on IRC, give me a ping

rootfs.py
Googulator: This part is pretty much nonfunctional at this point - it fails in populate_device_nodes, and even then, splitting it into 3 phases isn't really implemented (used to be sysa, sysc-pass1, sysc-pass2, but that's gone).

fosslinux (approved):

PR #358, opened 2023-12-20
Fix `--external-sources` and do not run `populate_device_nodes` in chroot mode

eduardosm:


eduardosm:
Not enough, still fails in some cases. I will open a new PR later.

PR #361, opened 2023-12-23
Fix bubblewrap support & CI workflows after simplify

Googulator:
Depends on https://github.com/fosslinux/live-bootstrap/pull/354 and https://github.com/fosslinux/live-bootstrap/pull/366


Googulator:
Seems like there are more checksum changes, and I don't currently understand why. I'll need to investigate this further.

Googulator:
The checksum changes are now resolved:

- kexec-linux had a wrong checksum in the current master already,
- perl-5.32.1 was affected by $USER leaking in due to no --clearenv support in Debian 10's version of bwrap,
- linux wasn't built reproducibly due to a missing timestamp override.

Googulator:
Source caching is now solved.

PR #367, opened 2023-12-26
CI: gzip intermediate stage tarballs to reduce network usage

Googulator:
Depends on #361. Split out into a separate pull request because #361 should be working already, and this change does have some risk of regression.


fosslinux:
determined that GH already gzips things for network transfer anyway, so this just adds unnecessary time onto CI.

PR #368, opened 2023-12-26
Rename tmpdir to target and always preserve it, dropping --preserve

Googulator:
There is nothing temporary about our "tmpdir" - its sole purpose is to contain the final product of the bootstrap process. Thus, removing it at the end of bootstrap amounts to doing the entire process for nothing.

To remedy this, --tmpdir is renamed --target, keeping the -t short form, and defaulting to "target" instead of "tmp" to make its purpose clearer. The --preserve option is removed, as the target is now always preserved.


stikonas (approved):

Googulator:
rebased on top of the gprof fix so we can actually get a green CI run

PR #370, opened 2023-12-26
Use anchors & aliases in GH actions

fosslinux:
Reduces code duplication

.. lets see if CI works


fosslinux:
... GH actions does not support anchors.

PR #371, opened 2023-12-26
Update `parts.rst` to remove remaining references to sysa/sysb/sysc

eduardosm:


parts.rst
```diff
@@ -782,7 +782,10 @@ curl 7.88.1
 ===========
 
 ``curl`` is used to download files using various protocols including HTTP and HTTPS.
-However, this first build does not support encrypted HTTPS yet.
+However, this first build does not support encrypted HTTPS yet. ``curl`` is built at
+this point because Linux must be running with support for threads (which is not the
+case if kernel bootstrapping is being used). curl requires musl with thread support,
+which was built at some point before Linux kernel.
```
fosslinux, line 788: This is not quite accurate, I think;

"curl requires musl with thread support, which was built at some point before Linux kernel." there should actually be a section between Linux kernel 4.9.10 and curl 7.88.1 that is this musl rebuild that gives musl thread support.

"Linux must be running with support for threads (which is not the case if kernel bootstrapping is being used)." At this point, Linux is guaranteed to have thread support, that's why we build it at this point.

Up to you if you want to fix this or I'll merge this then fix it.

eduardosm: Fixed (I think)

parts.rst
```diff
@@ -755,7 +755,7 @@ kexec-tools 2.0.22
 ``kexec`` is a utility for the Linux kernel that allows the re-execution of the
 Linux kernel without a manual restart from within a running system. It is a
 kind of soft-restart. It is only built for non-chroot mode, as we only use it
-in non-chroot mode. It is used to go into sysb/sysc.
+in non-chroot mode. It is used to boot the Linux kernel that will be built next.
```
Googulator, line 758: nitpick: "kexec" is only used to boot the final Linux kernel if the bootstrap kernel is also Linux (i.e. with --kernel).
In kernel bootstrap mode, the kernel we use up to this point is instead Fiwix, which means the regular kexec won't work, so the custom kexec-linux utility is used instead.

fosslinux (approved):
LGTM

PR #378, opened 2023-12-28
Fix "Bootstrapping" message being printed too late in bwrap runs

Googulator:
This should be printed at the beginning, but instead is only shown at the end, due to output buffering - force an explicit flush.


fosslinux (approved):
Not a behaviour I have noticed before, but I'll take your word for it, this is harmless :)

Googulator:
It was visible in a few CI runs, as well as almost every time I did a bwrap test in a Debian VM.

PR #381, opened 2023-12-28
 Create disk images without root 

eduardosm:
`mke2fs` has a `-d` option that allows to populate the newly created filesystem without needing to temporarily mount it. That allows to use `parted` and `mkfs.ext3` on regular files without needing root access.

I also fixed some issues I bumped into while testing (see commits).

### DIFF
--- a/lib/generator.py
+++ b/lib/generator.py
@@ -29,6 +29,7 @@ class Generator():
         self.external_sources = external_sources
         self.repo_path = repo_path
         self.source_manifest = self.get_source_manifest(not self.external_sources)
+        self.early_source_manifest = self.get_source_manifest(True)
         self.target_dir = None
         self.external_dir = None
 
@@ -159,9 +160,10 @@ class Generator():
 
     def distfiles(self):
         """Copy in distfiles"""
-        def copy_no_network_distfiles(out):
+        def copy_no_network_distfiles(out, early):
             # Note that "no disk" implies "no network" for kernel bootstrap mode
-            for file in self.source_manifest:
+            manifest = self.early_source_manifest if early else self.source_manifest
+            for file in manifest:
                 file = file[3].strip()
                 shutil.copy2(os.path.join(self.distfiles_dir, file),
                              os.path.join(out, file))
@@ -171,13 +173,13 @@ class Generator():
 
         if early_distfile_dir != main_distfile_dir:
             os.makedirs(early_distfile_dir, exist_ok=True)
-            copy_no_network_distfiles(early_distfile_dir)
+            copy_no_network_distfiles(early_distfile_dir, True)
 
         if self.external_sources:
             shutil.copytree(self.distfiles_dir, main_distfile_dir, dirs_exist_ok=True)
         else:
             os.mkdir(main_distfile_dir)
-            copy_no_network_distfiles(main_distfile_dir)
+            copy_no_network_distfiles(main_distfile_dir, False)
 
     @staticmethod
     def output_dir(srcfs_file, dirpath):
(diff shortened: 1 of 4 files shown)


Googulator:
A quick note: this should not affect probably the most common case of qemu or bare-metal with kernel bootstrap without --external-sources, which already proceeds without ever creating an FS image other than srcfs.

lib/generator.py
```diff
@@ -59,26 +59,16 @@ def prepare(self, target, using_kernel=False, kernel_bootstrap=False, target_siz
...
         elif using_kernel:
             self.target_dir = os.path.join(self.target_dir, 'disk')
-            target.add_disk("disk",
-                            filesystem="ext3",
-                            size=(target_size + "M") if target_size else "16G",
-                            bootable=True)
-            target.mount_disk("disk", "disk")
             self.external_dir = os.path.join(self.target_dir, 'external')
 
-        os.makedirs(self.external_dir, exist_ok=True)
```
Googulator, line 81: Why remove exist_ok here? I'm pretty sure this will break CI.

steps/jump/linux.sh
```diff
@@ -20,6 +20,7 @@ if [ "${KERNEL_BOOTSTRAP}" = True ]; then
     kexec-linux "/dev/ram1" "/boot/linux-4.9.10" "!$(command -v gen_init_cpio) /initramfs.list"
 else
     mkdir /etc
+    test -d "/proc" || (mkdir /proc && mount -t proc proc /proc)
```
Googulator, line 23: This will fail if /proc exists but isn't mounted.

I'd suggest:
```
mount | grep '/proc' &> /dev/null || (mkdir -p /proc; mount -t proc proc /proc)
```

Googulator (approved):
Other than these, LGTM.

lib/generator.py
eduardosm: Is there any situation where that directory can exist at that point? Previously, it would exist when the "external" image was mounted, but it is not longer the case.

steps/jump/linux.sh
eduardosm: Can `/proc` exist without being mounted? `populate_device_nodes` would need to be fixed too

lib/generator.py
Googulator: It exists at the beginning the latter phases of GitHub CI, which is split up into multiple phases to avoid hitting GitHub's time limits.

steps/jump/linux.sh
Googulator: Not right now, but this is needed if we ever want to boot back into a bootstrapped system, which is something I'm working on. (And yes, I will be applying the same fix to populate_device_nodes - in fact, I already have something similar in simplify-playground, although that one is outdated.)

eduardosm: Maybe we should call `populate_device_nodes` after building util-linux (which provides the `mount` command) and leave this jump script unmodified.

lib/generator.py
eduardosm: But the `prepare` function (the one modified here) is only called for the first step, later steps use the `reuse` function.

Googulator: Looks like I was actually mistaken about the purpose of this.

But it is indeed necessary - when building with kernel-bootstrap and external sources, /external will already exist when we get here.

eduardosm: I still don't see how can the directory exist.

> But it is indeed necessary - when building with kernel-bootstrap and external sources, /external will already exist when we get here.

That was true when the external image was mounted, which already created the directory. But now only the init directory will exist at that point.

steps/jump/linux.sh
eduardosm: Not longer relevant, since it has been fixed in https://github.com/fosslinux/live-bootstrap/pull/410

fosslinux (approved):

PR #382, opened 2023-12-29
Coreutils 9.4

fosslinux:


stikonas:
I fear that maintaining all those prepended checksums will be a bit painful...

fosslinux:
Removed the prepended checksums

stikonas (approved):

PR #385, opened 2024-01-01
Fix uninitialized variables in script-generator

Googulator:
Not sure if the Variable *last one is significant, but the Directive *last can be triggered by having the first directive in the manifest be a define.


stikonas (approved):

stikonas (changes requested):
This will need checksum changes?

Googulator:
> This will need checksum changes.

It appears script-generator isn't currently protected by a checksum. Fixing that should probably be a separate PR, though.

fosslinux (approved):

PR #386, opened 2024-01-01
 Support interactive issue resolution prompts

Googulator:
This adds a new flag, -i / --interactive, which enables opening a
Bash prompt whenever something goes wrong in the bootstrap. This is
highly useful when developing or debugging live-bootstrap, but it
needs to be off by default, for use in automated processes.

In the future, asking for variables at runtime could (and perhaps
should) also be gated behind this flag.

Depends on #385

### DIFF
--- a/rootfs.py
+++ b/rootfs.py
@@ -32,6 +32,7 @@ def create_configuration_file(args):
         config.write(f"UPDATE_CHECKSUMS={args.update_checksums}\n")
         config.write(f"JOBS={args.cores}\n")
         config.write(f"INTERNAL_CI={args.internal_ci or False}\n")
+        config.write(f"INTERACTIVE={args.interactive}\n")
         config.write(f"BARE_METAL={args.bare_metal}\n")
         if (args.bare_metal or args.qemu) and not args.kernel:
             if args.repo or args.external_sources:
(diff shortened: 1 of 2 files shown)


stikonas (approved):
Generally looks good to me. But do traps work in old bash? I remember having some issues early in the history of live-bootstrap.

Googulator:
Traps do work when used like this; the normal interactive prompt doesn't work, but this one does. I have hit it a few times during development already, and used the trap prompt successfully to fix the issue and continue bootstrapping.

Maybe EXIT traps don't work (that's what was used in live-bootstrap before the simplify refactor), but ERR ones do.

fosslinux (commented):
Perhaps change the variable name for using_bash... given it's not a boolean anymore.

Otherwise, OK with this implementation. In the future I think I'd like to support auto-interactivity in the case where bootstrap.cfg doesn't exist in the initial filesystem, but that would come later..

Googulator:
I was thinking "not using Bash" (0) -> "using Bash #1" -> "using Bash #2", not necessarily a boolean.

What would be a better variable name for you?

Googulator:
Just a quick heads up - I've just discovered a bug in this. When -i is given, checksum failures after switching to the first Bash don't trigger a drop to the trap as intended, but only print a warning. Fortunately, the fix should be easy ("set -E" needs to be set to propagate the trap to functions).

fosslinux:
> What would be a better variable name for you?

Perhaps `bash_rebuild`?
It's not a huge deal though.

Googulator:
"bash_rebuild" is slightly off semantically, as 0 means "no bash", 1 means "initial build of bash" (= no rebuilds), and 2 is "2nd build of bash") (= 1 rebuild). "bash_build" works, however.

Googulator:
Pushed the rename to bash_build.

Googulator:
Ouch, the "set -E" fix causes a trap in the autogen build

fosslinux (approved):

PR #387, opened 2024-01-01
Support swap files in qemu or bare-metal mode

Googulator:
This can help alleviate memory pressure when bootstrapping with high core counts.

### DIFF
--- a/steps/improve/swap.sh
+++ b/steps/improve/swap.sh
@@ -1,0 +1,19 @@
+
+#!/bin/sh
+#
+# SPDX-FileCopyrightText: 2024 Gábor Stefanik <netrolller.3d@gmail.com>
+#
+# SPDX-License-Identifier: GPL-3.0-or-later
+#
+# Set up swap
+
+. /steps/bootstrap.cfg
+. /steps/env
+
+if ! test -f /swapfile
+then
+    echo "Making swap..."
+    dd if=/dev/zero of=/swapfile bs=1M count=${SWAP_SIZE}
+    mkswap /swapfile
+fi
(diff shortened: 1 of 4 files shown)


stikonas:
Hmm, codewise looks good, but generally I think swapping during build is bad for performance. I usually find that reducing number of threads to avoid swapping is more helpful. Or did you observe different behaviour in live-bootstrap?

Googulator:
In the case of live-bootstrap, when going from 4 to 8 threads, the slight performance hit from occasionally hitting swap in GCC13 (the only place where 8 threads would otherwise go OOM) is far outweighed by the speed gain in Guile, especially on SSDs or other Flash storage. Haven't tested on an HDD though, results might be different there.

Also, Linux 4.9.10 is AFAIK already new enough to use swap in conjunction with disk caching even when there's no direct memory pressure, to increase performance.

steps/improve/swap.sh
```diff
@@ -0,0 +1,19 @@
...
+# SPDX-FileCopyrightText: 2024 Gábor Stefanik <netrolller.3d@gmail.com>
+#
+# SPDX-License-Identifier: GPL-3.0-or-later
+#
+# Set up swap
+
+. /steps/bootstrap.cfg
+. /steps/env
+
+if ! test -f /swap
```
fosslinux, line 13: nit, convention has become in mainstream Linux distributions to locate this at `/swapfile`. Although they are equally descriptive using the convention is probably better.

Googulator: Done.

fosslinux (approved):

PR #388, opened 2024-01-01
Restore multicore build support with kernel-bootstrap

Googulator:
This was removed as part of the simplify refactor, severely slowing down qemu and bare-metal builds. Restoring it brings us back to the same build times that we saw before the refactor.

### DIFF
--- a/rootfs.py
+++ b/rootfs.py
@@ -33,6 +33,7 @@ def create_configuration_file(args):
         config.write(f"UPDATE_CHECKSUMS={args.update_checksums}\n")
         config.write(f"JOBS={args.cores}\n")
         config.write(f"SWAP_SIZE={args.swap}\n")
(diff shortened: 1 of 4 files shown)


stikonas (approved):
`update_jobs` sounds a bit misleading. Maybe we can name it `update_build_job_count` or is that too long?

fosslinux (commented):
This is OK given the limitations of script-generator. Agree with stikonas that update_jobs is a bit misleading. Perhaps use_final_jobs would be better?

Googulator:
I settled for finalize_job_count in the end.

fosslinux (approved):

fosslinux:
ergh, conflicts...

Googulator:
Fixed.

PR #389, opened 2024-01-01
After bootstrap, drop to a shell if needed, then shut down cleanly

Googulator:


stikonas (approved):

fosslinux (commented):
For the `s` trigger, we can use `sync`.

I'm wondering if it might be better to read `/proc/mounts` and do the remounts ourselves... 

My concern is that `sleep 1` will be insufficient to complete each of these steps (there is no gurantee of that being long enough, particularly for slow storage media on fast CPUs, eg sd cards or USB drives).

Googulator:
Do we have the sync command working?

As for mounts, the only thing that really needs remounting is the root file system, as the rest is just tmpfs & "technical" file systems like procfs. Would "mount -o remount,ro /" be sufficient maybe?

fosslinux:
> Do we have the sync command working?

Yes, it is used in `steps/jump/linux.sh`.

> Would "mount -o remount,ro /" be sufficient maybe?

I think so. We only have that one disk based filesystem so it should be ok to do that. Technically we probably dont need it at all, but I'd prefer it to be safer...

Googulator:
Updated to use sync and mount instead of sysrq-trigger.

Googulator:
Testing the new implementation with explicit "mount" revealed that it unfortunately doesn't work. Apparently that's because the mount was made as "/newroot", before calling switch_root - and so "mount" won't recognize the root as a valid mount point.

Googulator:
This new version should work, although it's a bit ugly:

- Our version of mount expects to see valid mount point data in /etc/mtab, which is a real file on the root file system, rather than using /proc/mounts, which is maintained by the kernel. However, /etc/mtab will be copied to the real rootfs before the final switch_root, which makes it out of date. As a workaround, we cook /etc/mtab to rename /newroot to /, and drop /rootonly. This should be no longer needed if and when we upgrade the package that provides mount (util-linux?).
- Swap needs to be disabled before we can safely remount R/O. `swapoff -a` does the right thing regardless of whether or not a swapfile exists, but then it looks for /etc/fstab, finds it missing, and throws a gratuitous error, which needs to be suppressed using /dev/null and || true.
- GNU mount doesn't provide a real forcible remount option (`-f` is just "pretend it succeeded"). Work around this by first issuing a SysRq remount, and then only using `mount` to wait for it to finish.

fosslinux (approved):

PR #390, opened 2024-01-01
Use devtmpfs, and improve finalize_fhs.sh

Googulator:
With this, finalize_fhs.sh can be rerun as needed, e.g. when rebooting. Also, the preferred nameserver will persist after DHCP.

Thanks to devtmpfs, we no longer need to manage /dev once Linux is up.

### DIFF
--- a/steps/improve/finalize_fhs.sh
+++ b/steps/improve/finalize_fhs.sh
@@ -6,13 +6,22 @@
 #
 # Add the rest of the FHS that we will use and is not created pre-boot
 rm -rf /sbin /usr/sbin
+ln -s /usr/bin /usr/sbin
 for d in bin lib sbin; do
-    ln -s "usr/${d}" "/${d}"
+    ln -s "/usr/${d}" "/${d}" || true # these might exist if rerunning
 done
-mkdir -p /etc /run /var
-test -d /proc || (mkdir /proc && mount -t proc proc /proc)
-test -d /sys || (mkdir /sys && mount -t sysfs sysfs /sys)
+
+mkdir -p /etc /run /var/log /var/lock /var/spool /var/tmp /var/cache
+
+# can't use /dev/null before mounting /dev
+mount | grep '/dev' &> /junk || (mkdir -p /dev; mount -t devtmpfs none /dev)
+rm /junk &> /dev/null || true
+
+mount | grep '/proc' &> /dev/null || (mkdir -p /proc; mount -t proc proc /proc)
+mount | grep '/sys' &> /dev/null || (mkdir -p /sys; mount -t sysfs sysfs /sys)
 # Make /tmp a ramdisk (speeds up configure etc significantly)
-test -d /tmp || (mkdir /tmp && mount -t tmpfs tmpfs /tmp)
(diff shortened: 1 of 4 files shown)


stikonas (approved):

steps/improve/finalize_fhs.sh
```diff
@@ -6,13 +6,22 @@
...
 done
-mkdir -p /etc /run /var
-test -d /proc || (mkdir /proc && mount -t proc proc /proc)
-test -d /sys || (mkdir /sys && mount -t sysfs sysfs /sys)
+
+mkdir -p /etc /run /var/log /var/lock /var/spool /var/tmp /var/cache
+
+# can't use /dev/null before mounting /dev
+mount | grep '/dev' &> /junk || (mkdir -p /dev; mount -t devtmpfs none /dev)
+rm /junk &> /dev/null || true
```
fosslinux, line 18: Why `|| true` here?

steps/improve/finalize_fhs.sh
```diff
@@ -6,13 +6,22 @@
...
+
+mount | grep '/proc' &> /dev/null || (mkdir -p /proc; mount -t proc proc /proc)
+mount | grep '/sys' &> /dev/null || (mkdir -p /sys; mount -t sysfs sysfs /sys)
 # Make /tmp a ramdisk (speeds up configure etc significantly)
-test -d /tmp || (mkdir /tmp && mount -t tmpfs tmpfs /tmp)
+mount | grep '/tmp' &> /dev/null || (mkdir -p /tmp; mount -t tmpfs tmpfs /tmp)
+
 # Add /etc/resolv.conf
 echo 'nameserver 1.1.1.1' > /etc/resolv.conf
+echo 'nameserver 1.1.1.1' > /etc/resolv.conf.head
```
fosslinux, line 27: What's the purpose of this?

fosslinux: Oh, dhcpcd. We can do `echo 'nameserver 1.1.1.1' > /etc/resolv.conf > /etc/resolv.conf.head` for brevity

Googulator: Just tested that - it doesn't work.

`echo 'nameserver 1.1.1.1' | tee /etc/resolv.conf > /etc/resolv.conf.head` might, if we have `tee` available at this point.

steps/improve/finalize_fhs.sh
Googulator: Just defensive coding - if for any reason /junk doesn't get created, this way we don't needlessly destroy potentially several hours of bootstrap work.

steps/improve/finalize_fhs.sh
fosslinux: Ah - maybe not with the age of our bash.

fosslinux (approved):

steps/improve/finalize_fhs.sh
Googulator: It didn't work even in a rather modern bash I tested (5.1.16).

PR #391, opened 2024-01-01
Make the Linux kernel & Fiwix->Linux kexec more bare-metal-friendly

Googulator:
* Enable additional hardware drivers in Linux for better bare metal display, network & input device compatibility
* Disable ATA-over-Ethernet support, because it spams the network with unnecessary packets, is basically useless for bootstrapping, and may even be a security risk
* Increase Fiwix initrd size to 1280MB to fit a larger Linux build
* Make the Fiwix kexec size configurable the same way as initrd (and reduce from 280MB to 256MB which is sufficient in my testing)
* Use a more conservative memory map for Fiwix & Linux
* Boot Linux with consoleblank=0 on bare metal, so the build won't go blind after 5 minutes
* Support kexec-fiwix with interrupts disabled (will be useful later when builder-hex0 is updated)


fosslinux (approved):
Very nice, thank you!

I'm sure the future wil bring more changes with kernel config but this seems like a good start.

PR #392, opened 2024-01-02
Make move_disk work on bare metal

Googulator:
- Use -S32 -H64 --force to trick sfdisk into allowing MiB-aligned partitions on a drive that reports a CHS geometry
- Explicitly create partition at sector 2097152 (=1GiB)
- Force mkfs.ext4 to overwrite any existing filesystem it might find
- Wait up to 2 minutes for the disk to become readable (especially USB drives often show up with a delay)

Depends on #390

### DIFF
--- a/steps/jump/move_disk.sh
+++ b/steps/jump/move_disk.sh
@@ -4,7 +4,11 @@
 #
 # SPDX-License-Identifier: GPL-3.0-or-later
 
-set -ex
+set -e
+# mount might fail if /etc doesn't exist because of fstab and mtab
+mkdir -p /dev /etc
+mount -t devtmpfs none /dev &> /junk || true # no /dev/null yet
+rm /junk &> /dev/null || true
 
 # Create partition if it doesn't exist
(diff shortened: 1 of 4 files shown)


steps/jump/move_disk.sh
```diff
@@ -4,15 +4,31 @@
...
+        false
+    fi
+done
 
 # Create partition if it doesn't exist
 if [ $(($(stat -c "%Lr" "/dev/${DISK}") % 8)) -eq 0 ]; then
     echo "Creating partition table..."
-    echo ";" | sfdisk "/dev/${DISK}"
+    # Start at 1GiB, use -S32 -H64 to align to MiB rather than cylinder boundary
+    echo "2097152;" | sfdisk -uS -S32 -H64 --force "/dev/${DISK}"
```
fosslinux, line 28: I'm ok to leave this asis for now. But we might want an option later to disable this for use of smaller sized drives (although one could argue that we need enough space such that such a small sized drive is entirely impractical)

fosslinux (approved):
LGTM, all of these have been discussed before I think.

PR #397, opened 2024-01-06
Upgrade Fiwix to 1.5.0-lb1, pulling from upstream Mikaku repo.

### LINKED ISSUE #396: Update Fiwix to 1.5.0-lb1
All patches for Fiwix have been pulled into the primary Fiwix github repo.
We can stop using the rick-masters fork and pull directly from https://github.com/mikaku/Fiwix

There is one significant change from the fork that should be explained.

In the fork, the madvise system call was implemented as an empty stub just so madvise did not return an error when called. This was to workaround a bug in musl 1.1.24 which would call madvise internally and then incorrectly propagate an error from the syscall back to the application program (in errno), causing a failure. It was ultimately decided that it was more appropriate to patch musl than workaround the problem in Fiwix.
More details can be found [here](https://github.com/mikaku/Fiwix/issues/55).
The fix in the forthcoming PR mimics how musl 1.2.4 was fixed to ignore an error from madvise.

As a result of patching musl-1.1.24 many packages have new checksums in the forthcoming PR.

There are also changes to the way Fiwix is built.
Rather than passing many custom preprocessor definitions on the compiler command line, Fiwix now supports using custom header files to customize the kernel and so our definitions have been moved to these header files.

rick-masters:
Resolves #396


stikonas (approved):

stikonas (approved):

steps/musl-1.1.24/patches-pass3/madvise_preserve_errno.patch
```diff
@@ -0,0 +1,15 @@
+# SPDX-FileCopyrightText: 2023 Richard Masters <grick23@gmail.com>
```
stikonas, line 1: Other patches in this directory are stored as symlinks. Maybe this one should be as well?

rick-masters: Done.

PR #400, opened 2024-01-09
Upgrade autogen bootstrap tarball to fix trap issue

### LINKED ISSUE #399: Trap in autogen-5.18.16 in interactive mode (#386)
With the -i option from #386, autogen-5.18.16 hits a trap in bootstrap_autogen. In non-interactive mode, we somehow don't fail here, as if the relevant script was running without -e,

Googulator:
See https://github.com/schierlm/gnu-autogen-bootstrapping/pull/14

Fixes #399


Googulator:
Woohoo! Round numbered PR! :)

Googulator:
Just to clarify: this fixes an issue revealed by #386, but is in no way dependent on it; the issue was just hidden without #386.

stikonas:
Maybe let's see if we can get a new tag in the near future, so that we don't need to carry an extra patch.

Googulator:
Given that we're using a version that's already missing some fixes present in the upstream tree, where the last commit was over a year ago, I wouldn't hold my breath.

stikonas:
Oh, that's a patch for autogen, not for gnu-autogen-bootstrap. That's fine then.

fosslinux (approved):

Issue #355, opened 2023-12-18
Alternative tcc bootstrap path


cosinusoidally:
I've been working on an alternative bootstrap path for tcc (https://github.com/cosinusoidally/tcc_bootstrap_alt). I am able to go from a cut down version of stage0-posix-x86 (I only need cc_x86) all the way up to tcc-0.9.27. I hacked a version of live-bootstrap to use my alternative path (https://github.com/cosinusoidally/live-bootstrap/compare/master...cosinusoidally:live-bootstrap:tcc_bootstrap_alt). With my tweaks to live-bootstrap on my vm I was able to reduce the time to get to tcc-0.9.27 from around 30mins to around 3:45mins. It's also possible to use this to get to tcc-0.9.27 in < 512MB of RAM. The speed improvements are largely due to not building or using mescc (though I do still use meslibc).

I noticed that https://github.com/fosslinux/live-bootstrap/pull/334 has landed recently, so I'll probably refactor my live-bootstrap branch once that has stabalised and then raise a PR.

I was wondering what it would take to integrate this as an optionional bootstrap path for live-bootstrap? There are quite a few limitations to my approach (the major one being that it's i386 only) but it does have the advantage of being considerably faster.

fosslinux:
This may be a fairly good solution for #306 for 1.0, actually. Let me know if you hit any blockers with this.

PR #407, opened 2024-01-11
add alternative tcc bootstrap path

cosinusoidally:
This is a draft PR that adds my alternative bootstrap path for tcc https://github.com/fosslinux/live-bootstrap/issues/355 . I've marked this as a draft as it is probably not yet in a state suitable for merging (eg I updated the ci to unconditionally enable the flag  --tcc_bootstrap_alt but probably both the mes and tcc_bootstrap_alt path need testing automatically via ci). I also had some issues with nested conditionals in kaem that meant I needed to add some awkward combined conditionals (which also currently have bug that will break architectures other than x86/riscv). There may be a better way of doing that.


cosinusoidally:
I think the ci job on the pr is testing the ` --tcc_bootstrap_alt` route. I've kicked of a run of the mes route on my fork just to check I've not broken the standard path https://github.com/cosinusoidally/live-bootstrap/actions/runs/7489674266 (I have already tested a partial run of the standard path, but only up to tcc-0.9.27).

cosinusoidally:
Just pushed an updated version that computes sha sums as per our discussion on matrix. I had to split the tcc-0.9.26 hashes as some of them are only relevant to tcc-mes. I rebuilt the final version of tcc-0.9.26 a second time which caused the sha256sum to converge to the same value as the standard mes based bootstrap path. I couldn't get the same rebuild trick to work for tcc-boot0 and tcc-boot1 so I just ignore these hashes when the alt bootstrap path is enabled.

I've also added a separate ci step that tests my alt bootstrap path.

I've just noticed that my branch seems to break in qemu. It gets as far as the first call to cdrun here https://github.com/cosinusoidally/tcc_bootstrap_alt/blob/0.1/after.kaem#L20 and then crashes. I think it's failing around here https://github.com/cosinusoidally/tcc_bootstrap_alt/blob/0.1/tcc_js/cdrun.c#L48-L54 (based on the last thing printed to the terminal) . I'm guessing I may be hitting some issue in the chdir or execve syscall in builder-hex0

The command I used to generate my qemu image was something along the lines of:

```
./rootfs.py --bare-metal --target /tmp/build --tcc_bootstrap_alt
```

I used bare-metal as I was running the rootfs.py command inside a vm. I then grabbed the init.img file from /tmp/build and ran it inside a qemu vm on the host.

cosinusoidally:
My branch is well behind master now. In theory my https://github.com/cosinusoidally/live-bootstrap/tree/refs/heads/tcc_bootstrap_alt-refactor_nov24 branch will still work, but it's not in a mergeable state (actually it's not working https://github.com/cosinusoidally/live-bootstrap/actions/runs/18311814155/job/52141979882 I think https://files.bootstrapping.world/ might be down atm). I did create a variant of https://github.com/fosslinux/live-bootstrap/pull/524 which is much shorter simpler than the earlier approach I took in this PR, but my variant of the pnut PR is also not in a mergeable state. I may eventually create a new PR with my updated approach, but I might not get to that for a while.

PR #409, opened 2024-01-12
Fix panic due to after.sh prematurely quitting on shutdown

Googulator:
The sysrq shutdown trigger takes some time to fully shut down the system, during which init is expected to continue running. Since after.sh is the last step in our init, if it quits before shutdown is complete, Linux will panic with "Attempted to kill init".

Add an infinite loop after shutdown is issued via sysrq to prevent this.

### DIFF
--- a/steps/improve/after.sh
+++ b/steps/improve/after.sh
@@ -21,4 +21,5 @@ if [ "${CHROOT}" = False ]; then
     echo u > /proc/sysrq-trigger
     mount -o remount,ro /
     echo o > /proc/sysrq-trigger # power off
+    while true; do sleep 1; done
 fi


steps/improve/after.sh
```diff
@@ -21,4 +21,5 @@ if [ "${CHROOT}" = False ]; then
     echo u > /proc/sysrq-trigger
     mount -o remount,ro /
     echo o > /proc/sysrq-trigger # power off
+    while true; do true; done
```
eduardosm, line 24: There should be a sleep inside the loop body to avoid keeping the CPU busy.

Googulator: Good catch; this is far more serious on a single core system. Fixed.

fosslinux (approved):

PR #410, opened 2024-01-12
Populate devices so kexec can read /proc/iomem.

rick-masters:


Googulator:
I don't quite understand why this does anything at all, given that we run the same script a few steps earlier, but I guess it doesn't harm to re-run it, especially if it fixes something.

rick-masters:
> I don't quite understand why this does anything at all, given that we run the same script a few steps earlier, but I guess it doesn't harm to re-run it, especially if it fixes something.

When we run it earlier the `mount` command has not been built yet but after util-linux builds `mount` these commands will work in steps/improve/populate_device_nodes.sh:
```
if mount --version >/dev/null 2>&1; then
    test -d "/dev/shm" || (mkdir /dev/shm && mount -t tmpfs tmpfs /dev/shm)
    test -d "/proc" || (mkdir /proc && mount -t proc proc /proc)
fi
```

fosslinux (approved):

PR #414, opened 2024-01-21
Use builder-hex0 stages from submodule

Googulator:
This ends the reliance on the precompiled stage1 image in ootstrap-seeds, and the locally committed copy of the stage2 hex0 code.
Instead, both stages are taken from the new builder-hex0 submodule, with stage1 compiled by generator.py on the fly.
The local copy of stage2 is dropped.

This way, updating to a newer builder-hex0 becomes much easier: just update the submodule.

Note: The version of builder-hex0 thus pulled in is significantly newer than what we have been using until now, containing some much needed improvements to compatibility with real bare metal hardware & BIOSes. (USB flash drives may still not boot on many boards, due to BIOSes forbidding CHS access on HDD-emulated flash drives - a patch to builder-hex0 to use LBA in stage1 is coming soon.)


fosslinux (approved):
Thank you!

PR #415, opened 2024-01-22
Support reboot in non-chroot modes using grub and an init script

Googulator:
We use grub-2.06 as 2.12 has some packaging issues, and can't be built in live-bootstrap for now.

Depends on #418 (can be done separately, but then we'll have to remember to apply the changes to make_bootable.sh)


Googulator:
This has 2 bugs I've found so far:
- grub.cfg needs to reference $DISK, not $PHYSICAL
- our version of mount (util-linux) still relies on /etc/mtab, rather than /proc/mounts, so after reboot, we don't properly remount auxiliary filesystems - I'm still working on a fix for this

fosslinux (changes requested):
Just issues with pregened files in grub;

- there is a gnulib in `grub-core/lib/gnulib`
- see `autogen.sh` for a bunch of different regenerations that need to take place (unicode, libgcrypt, autogen-ish)


(Side note; the autogen here is WEIRRRD. Doesn't use proper autogen, rather uses some subset they wrote their own Python parser for a subset of autogen??)

Don't worry about the `contrib/` stuff, that's not really generated code...

Googulator:
Note also Stikonas's comments in https://github.com/fosslinux/live-bootstrap/pull/418 (they actually belong here)

Googulator:
> Just issues with pregened files in grub;
> 
>     * there is a gnulib in `grub-core/lib/gnulib`
> 
>     * see `autogen.sh` for a bunch of different regenerations that need to take place (unicode, libgcrypt, autogen-ish)
> 
> 
> (Side note; the autogen here is WEIRRRD. Doesn't use proper autogen, rather uses some subset they wrote their own Python parser for a subset of autogen??)
> 
> Don't worry about the `contrib/` stuff, that's not really generated code...

Should be resolved in the last force-push, but I'm still waiting for the build, to see if the checksum changes.

Googulator:
Looks like the checksum was unaffected.

Googulator:
Let's try again with a small fix to ensure reproducibility between 32-bit and 64-bit host kernels (otherwise a 32-bit userspace on a 64-bit kernel would be detected as "i686-pc-linux-muslx32" vs "i686-pc-linux-musl" on a 32-bit kernel).

steps/grub-2.06/pass1.sh
```diff
@@ -0,0 +1,35 @@
...
+
+    ./autogen.sh
+
+    cp -a INSTALL INSTALL.grub
+    autoreconf-2.69 -vif
+    mv INSTALL.grub INSTALL
+}
+
+src_configure() {
+    CFLAGS="-Wno-error" ./configure --prefix="${PREFIX}" --build=i686-pc-linux-musl
```
fosslinux, line 29: `--sbindir=${PREFIX}/bin`, please, because we aren't using separate `bin` and `sbin` in any other package

Googulator: Done, and rebased on current tip.

fosslinux (approved):

PR #418, opened 2024-01-23
Better interactive mode support

Googulator:
Improve interactive mode by:
- providing a proper TTY to the final bash prompt (fixes job control),
- spawning additional shells on virtual terminals to help monitor progress.

### DIFF
--- a/steps/manifest
+++ b/steps/manifest
@@ -54,6 +54,9 @@ build: bash-2.05b
 improve: setup_repo
 improve: update_env
 improve: merged_usr
+improve: populate_device_nodes
+define: CONSOLES = ( INTERACTIVE == True && CHROOT == False )
+improve: open_console ( CONSOLES == True )
 build: flex-2.5.11
 build: tcc-0.9.27
 improve: musl_libdir
@@ -80,7 +83,6 @@ build: perl5.004-05
 build: perl5.005-03
 build: perl-5.6.2
 uninstall: perl-5.000 perl-5.003 perl5.004-05 perl5.005-03
-improve: populate_device_nodes
 build: autoconf-2.52
 build: automake-1.6.3
 build: automake-1.6.3
@@ -128,11 +130,13 @@ jump: linux ( CHROOT == False )
 jump: move_disk ( KERNEL_BOOTSTRAP == True )
 improve: finalize_job_count
 improve: finalize_fhs
+improve: open_console ( CONSOLES == True )
 improve: swap ( SWAP_SIZE != 0 )
 build: musl-1.2.4
 build: curl-8.5.0
 improve: get_network ( CHROOT == False )
 build: bash-5.2.15
+improve: open_console ( CONSOLES == True )
 build: xz-5.4.1
 build: file-5.44
 build: libtool-2.4.7
--- a/steps/improve/after.sh
+++ b/steps/improve/after.sh
@@ -10,7 +10,7 @@
 . /steps/env
 
 if [ "${INTERACTIVE}" = True ]; then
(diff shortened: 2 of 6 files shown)


steps/grub-2.06/sources
```diff
@@ -0,0 +1 @@
+https://ftp.gnu.org/gnu/grub/grub-2.06.tar.xz b79ea44af91b93d17cd3fe80bdae6ed43770678a9a5ae192ccea803ebb657ee1
```
stikonas, line 1: I think we use mirrors.kernel.org for gnu stuff rather than the original gnu mirror

steps/improve/finalize_fhs.sh
```diff
@@ -23,6 +23,11 @@ mount | grep '/sys' &> /dev/null || (mkdir -p /sys; mount -t sysfs sysfs /sys)
 mount | grep '/tmp' &> /dev/null || (mkdir -p /tmp; mount -t tmpfs tmpfs /tmp)
 mount | grep '/dev/shm' &> /dev/null || (mkdir -p /dev/shm; mount -t tmpfs tmpfs /dev/shm)
 
+if [ "${CHROOT}" = False ]; then
+    rm /etc/mtab
+    ln -s /proc/mounts /etc/mtab
```
stikonas, line 28: I know we don't do this everywhere consistently, but generally long options in the scripts are more readable.

steps/improve/make_bootable.sh
```diff
@@ -0,0 +1,68 @@
+#!/bin/sh
+
+# SPDX-FileCopyrightText: 2024 Gábor Stefanik <netrolller.3d@gmail.com>
+#
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+# find the physical disk name
+PHYSICAL="${DISK}"
```
stikonas, line 8: I think quotes here are actually not necessary, it's assignment, so space expansion doesn't happen.

steps/grub-2.06/sources
Googulator: We do use ftp.gnu.org for gcc 10 and 13, and a few other components already. It seems to me more like mirrors.kernel.org is preferred for sources that rootfs.py needs to download (perhaps because of the - now fixed - throttling issue due to Requests' default UA string?).

stikonas (approved):

steps/grub-2.06/sources
stikonas: Might be just to be "nicer" and use a mirror and not because of throttling...

steps/improve/finalize_fhs.sh
Googulator: We use "ln -s" almost everywhere in live-bootstrap - in fact, I'm not sure if "-s" even has a standard long form.

steps/improve/make_bootable.sh
Googulator: This is actually #415

steps/grub-2.06/sources
Googulator: Anyway, this belongs to #415 and not here

steps/manifest
```diff
@@ -54,6 +54,8 @@ build: bash-2.05b
 improve: setup_repo
 improve: update_env
 improve: merged_usr
+improve: populate_device_nodes
+improve: open_console ( INTERACTIVE == True && CHROOT == False )
```
fosslinux, line 58: tiny nit: might be nice to do a `define: CONSOLES = ( INTERACTIVE...`

Googulator: Done and rebased.

Issue #379, opened 2023-12-28
chroot/bwrap fail to build automake-1.6.3 in WSL2 due to bash-2.05b globbing failure


Googulator:
Both chroot and bwrap builds fail in WSL2, because "rm -- configure Makefile.in */Makefile.in */*/Makefile.in aclocal.m4 automake.info*" fails to expand properly.

A simple testcase that shows this problem is as follows:
```
$ mkdir -p a/b
$ touch a/b/c
$ cd a
$ ./path/to/bad/bash -c 'echo */c'
*/c
$ bash -c 'echo */c' # system bash
b/c
```

bash-2.05b consistently behaves like this outside a chroot environment (even if it's run under "env -i"), but inside chroot/bwrap, it works fine - except on WSL2. IMO a proper fix would be one that makes it work outside a chroot environment as well.

Googulator:
This might be the fix: https://ftp.gnu.org/pub/gnu/bash/bash-2.05b-patches/bash205b-003

(not sure if our modern patch can read this format, but if not, it shouldn't be hard to convert to the modern format)

mid-kid:
I have the same issue. 
Here's the relevant log: http://0x0.st/HdnG.txt

Googulator:
As a workaround, upgrading the WSL2 kernel to linux-msft-wsl-6.6.y seems to avoid this issue.

This is not a fix, however - Bash 2.05b built using Autoconf works fine on the older kernel, the issue only occurs in Bash built via a custom makefile.

PR #419, opened 2024-01-24
Live system nochange

### LINKED ISSUE #377: src_install() shouldn't touch the live system
The following packages all touch the live system (adding, deleting or modifying files outside $DESTDIR) in their src_install:

- bash-2.05b (deletes old bash to avoid issues with tar later)
- bash-5.2.15 (ditto)
- coreutils-5.0 (directly creates /bin/pwd, bypassing $DESTDIR)
- flex-2.5.11 (removes lex from the live system)
- flex-2.6.4 (removes yacc from the live system)
- musl-1.1.24 (pass 3 only - removes ALL include files installed up to that point from /usr/include! This is fixed by #375 as it caused a checksum inconsistency.)
- perl5.004-05 (removes previous perl)
- perl5.005-03 (ditto)
- perl-5.6.2 (ditto)
- perl-5.10.1 (ditto)
- perl-5.32.1 (ditto)

All of these alterations to the live system will fail to be reproduced when these packages are installed from preseeded tarballs, rather than built in place.

If any of these changes is genuinely necessary, then we need to implement src_apply override support in helpers.sh, and move the relevant code there, so the proper FS changes will be made regardless of whether we're building or installing from preseed.

fosslinux:
a *bunch* of cleanups that make `diff -ru <(cd target; ls -R) <(cd target.bak; ls -R)` be nil, where `target` and `target.bak` have one run normally, and the other run with `--repo` from that run.

There are likely a couple of checksum changes I havent added yet, and there's still a bit of work to do for  `ls -lR` to be nil (permissions, etc)

Closes #377


seed/script-generator.c
```diff
@@ -212,12 +212,12 @@ Token *fill(Token *tok, Directive *directive, int type) {
...
 	/* logic = "("
-	 *         (name |
-	 *         (name "==" value) |
-         *         (name "!=" value) |
-	 *         (logic "||" logic) |
-	 *         (logic "&&" logic))
-	 *         ")"
+	 *	 (name |
+	 *	 (name "==" value) |
+     *	 (name "!=" value) |
```
Googulator, line 217: whitespace inconsistency

steps/bash-2.05b/pass2.sh
```diff
@@ -1,53 +0,0 @@
-# SPDX-FileCopyrightText: © 2023 fosslinux <fosslinux@aussies.space>
```
Googulator, line 1: We might need to bring this back to fix #379 (2.05b built via its normal build system doesn't have that issue)

steps/improve/clean_sources.sh
```diff
@@ -1,4 +1,4 @@
-# SPDX-FileCopyrightText: 2023 Eduardo Sánchez Muñoz <eduardosm-dev@e64.io>
+# SPDX-FileCopyrightText: 2024 fosslinux <fosslinux@aussies.space>
```
Googulator, line 1: it would be more polite to add your own copyright line after Eduardo's, rather than overwrite it

steps/improve/merged_usr.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+# SPDX-FileCopyrightText: 2023 fosslinux <fosslinux@aussies.space>
+#
+# SPDX-License-Identifier: GPL-3.0-or-later
+#
+# Add the rest of the FHS that we will use and is not created pre-boot
+ln -s bin /usr/sbin
```
Googulator, line 7: Do we still need to keep bin == sbin? In #415 by the time we build grub, /usr/sbin is its own directory again for some reason, so this doesn't seem to work as intended right now.

steps/manifest
```diff
@@ -178,10 +185,12 @@ build: python-3.3.7
 build: python-3.4.10
 build: python-3.8.16
 build: python-3.11.1
+uninstall: python-2.0.1 python-2.3.7 python-2.5.6 python-3.1.5 python-3.3.7 python-3.4.10 python-3.8.16
```
Googulator, line 188: I'd prefer to keep at least 2.5.6, to have a version of python 2.x available on the final system. (Of course, it would be better to build latest 2.7.x before 3.1.5.)

Googulator (approved):

steps/improve/clean_sources.sh
fosslinux: Oops, that's accidental!! Apologies to Eduardo :(

steps/improve/merged_usr.sh
fosslinux: This change fixes that, so that bin == sbin at the end of the bootstrap, at least on this branch

steps/manifest
fosslinux: Hm, is Python 2.x actually used anywhere nowadays? 

I'm not opposed to the idea, might just be unnecessary though

steps/bash-2.05b/pass2.sh
fosslinux: Good to note, I can either leave it existing in this PR, or the commit removing it can be `git revert`ed later?

Googulator: I'd rather leave it, as I'm almost done with the fix for #379, which depends on this script.

steps/manifest
fosslinux: Good point.

PR #420, opened 2024-01-24
Fix typo in script-generator causing wrong traps in new Bash

Googulator:
When renaming using_bash to bash_build, I accidentally removed the `+` sign from the assignment, causing `bash_build > 1` to never match, resulting in old-style Ctrl+D traps even in the new Bash.

This patch restores the intended `+=` behavior.


Googulator:
Confirmed on bare metal that this is indeed working as intended - 5.sh now has the correct trap line.

stikonas (approved):

PR #423, opened 2024-01-29
Add -F flag to mkfs to avoid prompt to proceed.

### LINKED ISSUE #422: Build with linux kernel prompts user to proceed
The commit to allow creating disk without root https://github.com/fosslinux/live-bootstrap/commit/679f73bbf126eab1ea8e73c10fc900f47c0ad064 started using mkfs.ext3 on a file image.

This triggers a prompt from mkfs.ext3:
```
Found a dos partition table in /home/ubuntu/live-bootstrap/target/disk.img
Proceed anyway? (y,N)
```

We need the `-F` flag because we are no longer running mkfs on a partition on a block special device.

From the man page:
```
       -F     Force mke2fs to create a filesystem, even if the specified device
 is not a partition on a block special device, or if  other  parameters do not make sense.
```

rick-masters:
Fixes #422

### DIFF
--- a/lib/generator.py
+++ b/lib/generator.py
@@ -98,7 +98,7 @@ class Generator():
                 mkfs_args = ['-d', os.path.join(target.path, 'external')]
                 target.add_disk("external", filesystem="ext3", mkfs_args=mkfs_args)
         elif using_kernel:
-            mkfs_args = ['-d', os.path.join(target.path, 'disk')]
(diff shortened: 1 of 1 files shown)


fosslinux:
In which situation does this occur?

fosslinux:
Oh, never mind, I see the issue.

fosslinux (approved):

PR #426, opened 2024-01-30
Opendoas (ie, sudo)

fosslinux:
This is significantly easier to build than sudo, which has a complicated build system and a few pregenerated files, but I was not nearly confident enough that I had found them all. We don't need all the multifaceted functions of sudo, really


parts.rst
```diff
@@ -1224,3 +1224,32 @@ to ensure the compiler is suitable for downstream consumption;
   really be handled by the libc, which is what most distributions do.
 * LTO now fully functions correctly, despite both the linker and the compiler
   being static binaries.
+
+<<<<<<< HEAD
```
stikonas, line 1228: merge conflict?

stikonas (approved):

PR #427, opened 2024-01-30
Add gzip 1.13

### LINKED ISSUE #369: Unpackaged, kaem-built gzip used in final bootstrapped image
Looking through the manifest for what happens to packages built using kaem, all but one of them either get rebuilt using bash (e.g. bzip2), replaced with a newer version (e.g. bash itself), or only needed early in the bootstrap, being replaced by completely different software (e.g. mes). The one exception is gzip, which is carried forward as an unpackaged binary all the way till the very end of the process.

This is obviously less than ideal. It would be nice if gzip was either rebuilt or upgraded at some point in the bootstrap.

fosslinux:
Closes #369


steps/gzip-1.13/sources
```diff
@@ -0,0 +1,2 @@
+https://ftp.gnu.org/gnu/gzip/gzip-1.13.tar.xz 7454eb6935db17c6655576c2e1b0fabefd38b4d0936e0f87f48cd062ce91a057
```
stikonas, line 1: most of the other downloads use mirrors.kernel.org mirror rather than pushing lots of traffic onto the main server...

stikonas:
Also add Fixes #https://github.com/fosslinux/live-bootstrap/issues/369 to commit message

stikonas (approved):

PR #429, opened 2024-01-30
Add gawk 5.3.0

fosslinux:


steps/gawk-5.3.0/sources
```diff
@@ -0,0 +1 @@
+https://ftp.gnu.org/gnu/gawk/gawk-5.3.0.tar.xz ca9c16d3d11d0ff8c69d79dc0b47267e1329a69b39b799895604ed447d3ca90b
```
stikonas, line 1: let's switch to the same mirror as elsewhere...

stikonas (approved):

steps/gawk-5.3.0/sources
fosslinux: Oh good point

stikonas (changes requested):
this is mismerged... There are various merge conflicts.

PR #432, opened 2024-02-05
Replace Fiwix filelist with parsing builder-hex0's FS structures

Googulator:
And while we're at it, use a more sustainable way of getting the address of the next file to be written to.

Since builder-hex0 doesn't distinguish directories from zero-length files, this has a limitation that it won't copy any zero-length files or empty directories. Only one such file is important for the bootstrap process (mes/config.h), which is recreated using an improve step.

Depends on https://github.com/ironmeld/builder-hex0/pull/11; the builder-hex0 commit referenced in the submodule is the one in that PR.


Googulator:
When combined with #434, this should work on bare metal, minus the missing checksum for m4-1.4.19 at the end (bypassable via --update-checksums).

Googulator:
Full test done - caught #435 upon rebooting, but otherwise fine.

stikonas (approved):
Good enough for now. Later we'll need to properly sort it our for UEFI bootstrap where reading files is different.

fosslinux (approved):
LGTM

Issue #394, opened 2024-01-02
script-generator isn't protected by a checksum


Googulator:
The script-generator utility lies right on the boundary between stage0-posix and live-bootstrap proper. As a result, it's protected by neither stage0's, nor live-bootstrap's checksum mechanism. This needs to be addressed for maximum trust & reproducibility.

PR #433, opened 2024-02-05
Improve pseudo-interactive prompts using the early bash

Googulator:
By wrapping $(cat) in an eval, redirections and other advanced syntax can now work in the early prompts.
Also, since "set -E" is broken is the early bash, fall back to using "set -e" and an EXIT trap, until we can upgrade to a bash version that already has working "set -E", or perhaps backport it to 2.05b.


Googulator:
No checksum update needed because of #394

fosslinux (approved):
LGTM

Googulator:
checksum-transcriber happens well before Bash, so it shouldn't be affected by Bash-related changes

Googulator:
Also, `-a amd64` isn't really supported yet; it's only for development - as such, breakages are likely (e.g. due to stage0 updates)

PR #434, opened 2024-02-06
M4 alternative update

fosslinux:
As m4 1.4.10 seems to have broken kernel bootstrap, revert that and rather add newest m4 to the end of the bootstrap.

(Untested under kernel bootstrap... but should be OK. Will test with all other changes)


Googulator:
Tested in kernel bootstrap too, combined with #432 and #435, it all works. The 1.4.19 checksum is missing, though.

stikonas (approved):
Looks good apart from checksum that @Googulator already mentioned

PR #436, opened 2024-02-09
Support early xz/lzma decompression, and use it wherever possible

Googulator:


stikonas (approved):

steps/make-3.82/sources
```diff
@@ -1 +1 @@
-https://mirrors.kernel.org/gnu/make/make-3.82.tar.bz2 e2c1a73f179c40c71e2fe8abf8a8a0688b8499538512984da4a76958d0402966
+http://mirror.clarkson.edu/slitaz/sources/packages/m/make-3.82.tar.lzma 9475ef76dd33c319aa8a48d8bca28b2312b90b5e59c2b0b05b206c3f50614ee2
```
stikonas, line 1: oh, this one is a bit strange. Looks like the only source for lzma tarball? Did they repackage it themselves?
@fosslinux what do you think about this one?

Googulator: Right, I wasn't able to find any other source for this tarball.

fosslinux: Hm, I think I'd prefer to use the more widespread bz2, not sure it's worth saving 252KB, but happy to hear other options.

fosslinux (approved):
I'm cool with this :)

PR #437, opened 2024-02-12
 Use e820 memory map from BIOS, if available

Googulator:
This extends make_fiwix_initrd and kexec-fiwix to support command line parameters, instead of hardcoding relevant values within the C sources. This way, it becomes possible to alter e.g. ramdisk size without affecting checksums.

While we're at it, also support loading a memory map from a file populated by builder-hex0 based on the BIOS E820 memory map.

Depends on https://github.com/ironmeld/builder-hex0/pull/12

### DIFF
--- a/steps/jump/fiwix.sh
+++ b/steps/jump/fiwix.sh
@@ -7,7 +7,11 @@
 set -ex
 
 # Build the ext2 image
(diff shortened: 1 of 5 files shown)


fosslinux (approved):
Nice!!!

Issue #443, opened 2024-02-18
Enable KVM support in the Linux kernel


Googulator:
From #442: "KVM, unfortunately, doesn't compile with our current toolchain; I tried to patch it out, with no luck. It causes undefined reference to `__compiletime_assert___COUNTER__' errors at the final link, and unlike drm_edid, this doesn't seem to be patchable just by removing BUILD_BUG_ONs."

This issue is for tracking and solving that compile error.

PR #444, opened 2024-02-18
Update Linux kernel configuration & patches

Googulator:
This updates the Linux kernel configuration, removing irrelevant networking-related components, and switching framebuffer drivers to the DRM-based ones. With this, we can finally bootstrap on systems with newer NVIDIA cards, which would green screen with the nvidiafb driver.

KVM is still disabled, see #443 - VIRTUALIZATION is also disabled, as it's useless without KVM or any other suboption being enabled.

Additionally, we can now drop the bad-asm patch, since our newer binutils has no problem supporting that syntax, and it doesn't look quite innocuous to me, removing an offset from a memory access.

The kernel is now built in 2 stages: first, we build vmlinux only, then, after cleaning up any intermediate .o files (except the ones needed to build efistub), we convert it to a bzImage. This required some creative use of the -o option to convince Make not to rebuild all of the .o files we've just deleted as dependencies.


stikonas (approved):

fosslinux (approved):
LGTM

PR #445, opened 2024-02-20
Check "substitutes"

fosslinux:
I think it is likely useful to check that files we get from downstream mirrors are equivalent to a "primary" mirror, particularly when said files do not have the same checksum for whatever reason - recompression, etc.

This script checks them automatically as part of CI.


stikonas (approved):

stikonas:
Generally looks good but need to fix linters.

PR #455, opened 2024-04-13
Upgrade kernel to 4.14.341-openela

### LINKED ISSUE #452: Upgrade kernel to 4.14.341-openela
https://github.com/openela/kernel-lts is a version of the v4.14 Linux kernel, maintained by https://openela.org/ on top of the last kernel.org release, ostensibly to the same standards as the now-discontinued kernel.org point releases.

Unlike kernel.org, OpenELA provides no XZ tarballs, only GZ ones, which would result in an unreasonable srcfs footprint growth. However, the diff from the last kernel.org release to the latest OpenELA one (excluding the ".elts" metadata directory) is less than 500KiB, which is much more reasonable to include.

Diffs can be generated using:
```
git clone https://github.com/openela/kernel-lts/
cd kernel-lts/
git diff --no-prefix v4.14.336 v4.14.341-openela -- . :^.elts > 4.14.341-openela.diff
git diff --no-prefix v4.14.336 v4.14.341-openela -- scripts include arch/x86/include arch/x86/entry > 4.14.341-openela-headers.diff
```

Dropping these into the patches directories in linux & linux-headers provides an easy path to upgrade to these still-supported kernels.

OpenELA intends to support this fork at least until the end of 2024.

Googulator:
This is a continuation of the 4.14 series by the Open Enterprise Linux Alliance, ostensibly to the same maintenance standards as the now discontinued kernel.org LTS.

This code is primarily offered as a Git repository, with gz tarballs also available, but no xz versions. Switching back to a gzipped version of the kernel source code would introduce too much srcfs growth, so we use the last kernel.org tar.xz release (4.14.336), and apply OpenELA's changes using a patch generated from the Git repository.

Fixes #452


stikonas (approved):

Googulator:
Tested also on bare metal, both initial bootstrap and reboot work fine.

PR #457, opened 2024-04-14
Move heirloom lex and flex after musl, replace heirloom yacc with byacc

### LINKED ISSUE #456: Use byacc instead of heirloom yacc for Bash's .y files (avoid linking heirloom-devtools to GPLed meslibc)
Normally, bash-2.05b requires bison to rebuild its .y files. Because there's no easy way to build bison without already having a working Bash, we currently use yacc from heirloom-devtools instead - unfortunately this version of yacc is CDDL, while meslibc (the only libc available then) is GPL, which are incompatible licenses.

Another yacc implementation, byacc, is in the public domain, and comes with a simple makefile - unfortunately this makefile was abandoned a while ago, and doesn't build modern versions of byacc, which are instead meant to built via autotools. Rather than searching for an older version of byacc, I chose to update the makefile, resulting in https://gist.github.com/Googulator/9eee5c0ccd10fd819bf937b4f650b52f which is able to build a working byacc on my modern Linux system.

With this, hopefully we can go from:
```
build: heirloom-devtools-070527
build: bash-2.05b
improve: setup_repo
improve: update_env
improve: merged_usr
improve: populate_device_nodes
define: CONSOLES = ( INTERACTIVE == True && CHROOT == False )
improve: open_console ( CONSOLES == True )
build: flex-2.5.11
build: tcc-0.9.27
improve: musl_libdir
build: musl-1.1.24
```
to:
```
build: byacc-20240109
build: bash-2.05b
improve: setup_repo
improve: update_env
improve: merged_usr
improve: populate_device_nodes
define: CONSOLES = ( INTERACTIVE == True && CHROOT == False )
improve: open_console ( CONSOLES == True )
build: tcc-0.9.27
improve: musl_libdir
build: musl-1.1.24
build: heirloom-devtools-070527
build: flex-2.5.11
```
hence only linking heirloom-devtools to musl, whose license is compatible with CDDL.

Googulator:
This way, heirloom-devtools is only linked to a libc with a CDDL-compatible license, making it fully redistributable. Heirloom yacc doesn't work properly with musl libc, but luckily byacc can do its job just fine, so we only need heirloom lex now.

Thanks to the more complete libc, and bash being available, heirloom's bootstrap is significantly simplified.

Fixes #456


stikonas (approved):
Looks good. Let's see how to fix reuse linter...

stikonas (approved):

PR #468, opened 2024-05-12
Add interactive "configurator"

fosslinux:
This is a more UX-friendly glorified bootstrap.cfg editor. Somewhat inspired by kconfig.

Mostly complete, but WIP still

- Does NOT run if INTERACTIVE=False. (Should we introduce a different environment variable for this?)
- No current support for conditional selections of options based upon other options.

Known bugs:

- <strike>validation of string inputs does not currently work</strike>
- ?

Depends on [oriansj/M2libc#](https://github.com/oriansj/M2libc/pull/59)


Googulator:
Did you test if this actually works with kernel bootstrap? At that point, we're still running on builder-hex0, which may not support keyboard input.

fosslinux:
I'm a fair bit more happy with the state of this now. It works under builder-hex0, at a later stage, and doesn't run if you do use rootfs.py to generate a complete bootstrap.cfg file. There is a separate variable toggling use of configurator.

I believe that perhaps this may have pushed Fiwix over the ramdisk size as far as I can tell, but retesting that now. Planning to do a little bit more testing but would appreciate re-reviews at this point.

Googulator (approved):
Looking good now, thanks.

seed/configurator.c
```diff
@@ -0,0 +1,784 @@
...
+	 * Check if we are NOT running under fiwix or linux.
+	 * If we are not, and need configuration to occur, then we presume that
+	 * we will not be able to get any input from the user.
+	 */
+	struct utsname *kernel = calloc(1, sizeof(struct utsname));
+	uname(kernel);
+	if (kernel->sysname == NULL) {
+		no_input(head);
+		return 0;
+	} else if (strcmp(kernel->sysname, "Linux") != 0 && strcmp(kernel->sysname, "Fiwix") != 0) {
```
Googulator, line 715: I wonder if there's a more future-proof way of detecting builder-hex0, that won't come back to bite us when we try to bootstrap on e.g. Hurd or macOS.

fosslinux: I looked briefly for any key features of builder-hex0 that make it detectable, but it appears to be fairly silent on its existence - which kind of is a good thing I guess!

Either way, we'll have to modify the checks at some point; either for stage0-uefi, or for other kernels live-bootstrap can be run on.

PR #469, opened 2024-05-22
Bootstrap Guix on top of live-bootstrap

Googulator:


stikonas:
From what I've seen before, it's musl's `./configure` script that for some reason replaces `/dev/null` block device with a file. Probably we are triggering some bug.

vxtls:
WIP https://github.com/vxtls/live-bootstrap/tree/live-bootstrap-to-guix

PR #472, opened 2024-06-21
Update URL for openssl-3.0.13.tar.gz

pyy:


Googulator:
OpenSSL is what enables access to HTTPS URLs, so we can't pull OpenSSL via HTTPS.

stikonas:
Also target branch is incorrect, it should go to 1.0 branch.

stikonas:
> > OpenSSL is what enables access to HTTPS URLs, so we can't pull OpenSSL via HTTPS.
> 
> But the old link is invalid. Should we use third-party repos, like some linux repos to get openssl-3.0.13.tar.gz?

I think third-party mirror should be fine. See if there is anything with 3.0.13.

Googulator:
See also: https://github.com/openssl/openssl/issues/24706

stikonas:
On #bootstrappable BearSSL was mentioned, longer term it might be worth looking into including that into pre-network part of bootstrap. Though perhaps for short term we could switch to HTTP mirror.

PR #478, opened 2024-07-31
Enable _ctypes and dynamic linking for python

mid-kid:


Googulator:
Looking good, thanks. (This is actually one of the things needed for bootstrapping Guix.)

PR #479, opened 2024-09-11
Allow the user to use wrap as part of the bootstrap process

MaxHearnden:
This adds support for wrap in three situations:
1. Using `rootfs.py -w`
2. In a prepared environment containing distfiles, seed and steps (as sub directories)
3. In a clone of the whole git repository with distfiles (Although it does overwrite tracked files within the git repository)

For 2 and 3, wrap can be used by copying `seed/after-wrap.kaem` to `seed/stage0-posix/after.kaem` (symlinking works as well) and running `bootstrap-seeds/POSIX/${ARCH}/kaem-optional-seed` from within the `seed/stage0-posix` directory

This replaces https://github.com/fosslinux/live-bootstrap/pull/339


Googulator:
I suggest adding a GitHub CI workflow using the -w option.

Also, mind the Python linter.

MaxHearnden:
Done, just waiting for the CI to finish

fosslinux:
If you're comfortable with it, could you use a matrix for the Github workflow, instead of copy-pasting the workflow file? https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/running-variations-of-jobs-in-a-workflow

If you're not comfortable I can do that.

MaxHearnden:
I've added a test matrix, would you like me to use separate step 1 and step 2 artifacts for bwrap and wrap.

Googulator:
Yes, that would be preferable.

MaxHearnden:
It now uses separate artifacts

lib/generator.py
```diff
@@ -134,12 +140,17 @@ def stage0_posix(self, kernel_bootstrap=False):
...
         """Copy in extra seed files"""
         seed_dir = os.path.join(self.git_dir, 'seed')
         for entry in os.listdir(seed_dir):
             if os.path.isfile(os.path.join(seed_dir, entry)):
                 shutil.copy2(os.path.join(seed_dir, entry), os.path.join(self.target_dir, entry))
+        if wrap:
+            shutil.copy2(os.path.join(seed_dir, 'after-wrap.kaem'),
+                         os.path.join(self.target_dir, 'after.kaem'))
+            shutil.copy2(os.path.join(seed_dir, 'after.kaem'),
+                         os.path.join(self.target_dir, 'after-wrapped.kaem'))
```
fosslinux, line 153: I tend to try to avoid having conditional logic in `generator.py`. Is there some reason this needs to be here, and can't occur within the live-bootstrap environment?

Ie, there would be conditional logic in after.kaem for wrap mode, by adding in some kind of `WRAP=True` into `bootstrap.cfg`, rather than replacing the kaem files like this.

seed/after.kaem
```diff
@@ -11,5 +11,18 @@ set -ex
 
 PATH=/${ARCH_DIR}/bin
 
-catm seed-full.kaem /steps/bootstrap.cfg /steps/env seed.kaem
+if catm seed-full.kaem /steps/bootstrap.cfg /steps/env seed.kaem; then
+else
+  replace --file /steps/env --output /steps/env --match-on /external/distfiles --replace-with /distfiles
```
fosslinux, line 16: Why is `/distfiles` being reintroduced?

seed/wrap-bootstrap.cfg
fosslinux, line 1: This file should probably not exist, and should be generated in `rootfs.py` instead. (I'm not too sure why this file exists anyway. Why can't someone have `BUILD_KERNELS` enabled under wrap mode, for instance?)

steps/improve/merged_usr.sh
```diff
@@ -6,5 +6,9 @@
 # Add the rest of the FHS that we will use and is not created pre-boot
 ln -s bin /usr/sbin
 for d in bin lib sbin; do
+    if [ -d "/${d}" ] && ! [ -L "/${d}" ]; then
+        # Move the non symlink directory out of the way
+        mv "/${d}" "/${d}-saved"
+    fi
```
fosslinux, line 12: I'm reading the commit description where this was introduced but still cannot figure out why this was added.

If I'm understanding the commit, my best guess is because the entire git repository is being copied into `target`, but I'm not sure why that is desirable either (not a supported running method)

fosslinux (changes requested):
Thanks for the contribution! There's a few structural things that I'd like to see changed a bit, they go against convention that we've been using. Particularly, we try to make the bootstrap flow as mode-agnostic as possible, and simply wrap anything specific to one mode in an if block, but the way it's currently structured is difficult to follow.

seed/wrap-bootstrap.cfg
MaxHearnden: That file is only used if rootfs.py isn't being used and is used in combination with the configurator, as such I put reasonable defaults in.

seed/after.kaem
Googulator: I believe the idea here is to just run the bootstrap directly from the checkout directory, bypassing rootfs.py or any manual preparation. Since distfiles are downloaded to ${checkout_dir}/distfiles, with no parent named "externals" guaranteed to exist, /external/distfiles will fail in this scenario, no matter what directory you map as the root.

The alternative is to download distfiles to ${checkout_dir}/external/distfiles (which I would somewhat prefer to this solution), or to just not support this use case at all, no matter how compelling it seems.

steps/improve/merged_usr.sh
Googulator: Again, I suspect this is related to running without a target directory, i.e. straight from the checkout directory.

PR #484, opened 2024-12-17
Make Python optional & Modernise Github Actions

fosslinux:
These two are fairly interlinked so I just included them together.

**CI Must run to completion**

### DIFF
--- a/rootfs.py
+++ b/rootfs.py
@@ -220,13 +220,10 @@ print(shutil.which('chroot'))
 
     elif args.bwrap:
         init = '/init'
-        if not args.internal_ci or args.internal_ci == "pass1":
-            generator.prepare(target, using_kernel=False)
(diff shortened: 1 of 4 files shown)


rootfs.py
```diff
@@ -101,6 +103,8 @@ def main():
     parser.add_argument("--internal-ci", help="INTERNAL for github CI")
     parser.add_argument("-s", "--swap", help="Swap space to allocate in Linux",
                         default=0)
+    parser.add_argument("-x", "--extras", help="Comma-separated list of extra bootstraps to run (python)",
```
stikonas, line 106: I think `choices = ['python']` with `nargs="+"` would work better. You would not need to do the checks bellow then.

stikonas (changes requested):
Do we need to enforce all extras when calculate checksum mode is activated

PR #487, opened 2024-12-21
CI fixes

fosslinux:


stikonas (approved):

.github/workflows/bwrap.apparmor
```diff
@@ -0,0 +1,13 @@
+# SPDX-FileCopyrightText: 2024 Russell Coker <russell@coker.com.au>
+# SPDX-License-Identifier: CC-BY-NC-SA-3.0
```
Googulator, line 2: Assuming this is even copyrightable, it might be worth contacting the author about a better license for this file, as NC licenses could be problematic for things like GPL compatibility. (Even though this is only used in the CI.)

stikonas: The same thing here but with a different binary (firefox):

https://support.mozilla.org/en-US/kb/install-firefox-linux

Issue #485, opened 2024-12-18
RFC: Source distribution


fosslinux:
I think we need to reconsider our model for the distribution of input tarballs/distfiles into live-bootstrap.

## State of play

We have three "distinct"-ish sections of the bootstrap in this area, each of which have been treated with somewhat different requirements.
- pre-networking. Before networking is available, all distfiles must be pre-loaded onto the system.
- pre-SSL. Once networking is available, we immediately build curl, so we have the option (if `--external-sources` is off), to download sources *within* the bootstrapped system. However, we cannot access HTTPS sites at this point, as we don't have SSL support. Therefore, all distfiles in this stage must be available over HTTP (non-SSL).
- post-SSL. At this point, we have curl with SSL support, so we can get distfiles over HTTPS.

And we are currently using two ways to get distfiles:
- HTTP
- HTTPS

Note: some distfiles are effectively an endpoint running on-demand, or serving a cached, `git archive`.

Here are some "de-facto" rules we have been using;
- HTTP and HTTPS is allowed for pre-networking and post-SSL stages.
- HTTP only is allowed for pre-SSL stages.
- there is a non-#bootstrappable/bootstrappable.world source available for each distfile.
  - this has proved particularly challenging in the pre-SSL stage, where there are often few HTTP-only sites available, and for Git snapshots, which are quite unreliable (currently, Gnulib is a problem)

### Proposal: Do not require a HTTP-only, non-#bootstrappable source for each distfile in the pre-SSL stage.

**Currently:** We need an upstream source, or a mirror, or archive.org, hosted on HTTP, for every distfile in the pre-SSL stage.
**Suggestion:** Host them ourselves on a HTTP-enabled server. This is OK, because it will have the same checksum as the upstream anyways. Furthermore, once SSL is available, it is easy to check the file from upstream also matches the checksum.

Problems:
- We control both the checksum and the distfile, so malicious changes could be easily slipped in.
  - Mitigation: It is easy to check that the distfile is equivalent, using checksums.
  - Mitigation: See proposal below regarding mirror network.

### Proposal: Create git snapshots ourselves using `git archive` and distribute them ourselves, instead of using Git snapshots from cgit/gitweb/GitHub/similar.

**Currently:** If we need a particular Git commit, we download a snapshot of it from something like cgit, gitweb or GitHub. These tend to be unreliable or just randomly disappear (see Gnulib). Further, no-one is checking that the files are the same in the Git repository as they are in the generated snapshot.
[...]

PR #493, opened 2025-01-04
Mirrors

fosslinux:
Here's a first attempt at a mirrors network as per #485.

This is *quite basic*, and intended to be iterated upon to possibly add things like IPFS/torrents, mirror discovery, etc.

At this point in time, this PR only attempts to implement the following model:
- mirrors are given by a static list as a configuration variable
- each mirror has a copy of all distfiles, and creates `git archive` snapshots

The "sources" format has a change: there is now a Git line supported, which has the format:

```
git://<Git url> <url to a Git archive snapshot, if one exists, or else an underscore> <checksum of the Git archive snapshot> [filename]
```

`download-distfiles.sh`, `rootfs.py` and `helpers.sh` have been updated to reflect this.

`mirror.sh` is an implementation of a mirror creation/maintenance tool, it is intentionally written in **POSIX sh** (not bash, like we use for most other things), because platform diversity should be encouraged. I would also encourage other mirror implementations to be written eventually.

Alongside that, in this PR I have accordingly swapped out a number of distfiles to get us "closer to upstream";

- all gnulib snapshots have been changed to `git archive` snapshots
- all GitHub *snapshots* (not release artifacts, which could be non-dynamically generated tarballs) have been changed to `git archive` snapshots; namely, gnu-autogen-bootstrapping in autogen, guile-psyntax-bootstrapping in guile, kexec-tools
- other Savannah snapshots, namely Autogen and coreutils, have been changed to `git archive` snapshots
- bash-2.05b, bc-1.07.1, bzip2-1.0.8, ed-1.4 have all had their tarballs moved back to upstream, rather than recompressed versions created by various distributions
- gcc-10.4.0 and gcc-13.1.0 have moved to mirrors.kernel.org for consistency

As of yet unresolved questions;

- what do we do about CI now? should it use the main mirror network? or should it use a special CI mirror?


mirror.sh
```diff
@@ -0,0 +1,182 @@
+#!/bin/sh
+# This script is intentionally written in POSIX sh to be cross-platform.
+
+set -e
```
stikonas, line 4: add `-o pipefail`, else -e might be insufficient.

fosslinux: Isn't `-o pipefail` a BASH-ism?

stikonas: Oh indeed...

DEVEL.md
```diff
@@ -84,3 +84,43 @@ libraries that are compiled and installed.
...
+Git protocol path, except for GitHub, in which `git://` is automagically
+changed to `https://`.
+
+The URL to HTTP snapshot may be `_` (a single underscore) if no HTTP snapshot
+exists. In this case, the filename is compulsory. The checksum is of the Git
+snapshot, generated with `git archive`. See `mirror.sh` for more detailed rules.
+
+Some specific helpful things:
+- prefer `.tar.gz`
+- For GitHub, all snpashots are generated with long commit IDs, so use long
```
stikonas, line 113: `snpashots` -> `snapshots`

download-distfiles.sh
```diff
@@ -4,26 +4,46 @@
...
 
+# Optional arguments: Mirrors
+
 download_source() {
     local distfiles="${1}"
     local url="${2}"
-    local checksum="${3}"
-    local fname="${4}"
+    shift 2
+    if [[ "${url}" == git://* ]]; then
```
stikonas, line 13: I think double quotes are not needed for `[[`

download-distfiles.sh
```diff
@@ -4,26 +4,46 @@
...
+        url="${1}"
```
stikonas, line 14: I think double quotes here and further below on `local` assignment lines are also not required, since this is assignment and not use of the variable.

lib/simple_mirror.py
```diff
@@ -0,0 +1,23 @@
...
+# SPDX-FileCopyrightText: 2025 fosslinux <fosslinux@aussies.space>
+
+import http.server
+import socketserver
+
+class SimpleMirror(socketserver.TCPServer):
+    """Simple HTTP mirror from a directory"""
+    def __init__(self, directory: str):
+        self.directory = directory
+        super().__init__(("127.0.0.1", 0), self._handler)
```
stikonas, line 15: Perhaps `localhost` would be better. That will also be IPv6 compatible.

download-distfiles.sh
fosslinux: Oh, this is true.

lib/simple_mirror.py
fosslinux: Good idea.

mirror.sh
stikonas: So on #bootstrappable it was pointed out that it is now part of POSIX.1-2024.

mirror.sh
```diff
@@ -0,0 +1,185 @@
...
+if [ "${state}" = "" ]; then
+    state="${PWD}/mirrorstate"
+fi
+state="$(realpath "${state}")"
+
+#######
+
+# Download a HTTP file
+download_file() {
+    url="${1}"
```
stikonas, line 37: more of the optional quotes on assignment in this file

steps/manifest
```diff
@@ -84,7 +84,7 @@ build: perl-5.003
 build: perl5.004-05
 build: perl5.005-03
 build: perl-5.6.2
-uninstall: perl-5.000 perl-5.003 perl5.004-05 perl5.005-03
+uninstall: perl-5.000 perl5.004-05 perl5.005-03
```
stikonas, line 87: is this change intentional?

stikonas (approved):
A couple of minor comments/questions left but otherwise looks fine.

mirror.sh
```diff
@@ -0,0 +1,185 @@
...
+        # All git repositories we already use remove "v"s from the beginning
+        # of branch/tag names in the tarball prefix
+        if echo "${reference}" | grep -Eq "^v[0-9]"; then
+            prefix_ref=$(echo "${reference}" | sed "s/^v//")
+        fi
+        prefix=$(basename "${repo}" | sed "s/.git$//")-${prefix_ref}
+
+        (
+            cd "${outdir}" || exit
+            git config tar.tar.gz.command gzip
```
stikonas, line 164: There was some discussion now on #bootstrappable. Looks like `gzip` is not deterministic. At least on files. `gzip --no-name` might be better. Or perhaps internal git implementation?

stikonas (approved):

PR #496, opened 2025-02-03
GCC fixes: Updates & Regenerations

fosslinux:
As a part of linting each distfile used in live-bootstrap using problematic-source, I of course came across some (many?) things in GCC. Also updated GCC where relevant.

Summary:

- every GCC version had a number of *unused* generated files in the tree
  - I want to document these anyway, but the question is, should we remove them anyway? There are a lot of them. Is the clutter to the scripts worth it?
  - it does avoid an xz-style attack

No huge misses were found, but there were a lot of small generated files found here and there.
The only problematic one was `decDPD.h`, which is actually used and had no information about how it was made. After a bit of searching, I found the original generating script in an obsolete language OORexx, and ported it to C.


stikonas (approved):

steps/gcc-4.7.4/files/decDPD_generate.c
```diff
@@ -0,0 +1,228 @@
...
+
+#include <stdio.h>
+#include <stdint.h>
+#include <stdbool.h>
+
+/* 
+ * Creates decDPD.h.
+ * Based upon the algorithm given on;
+ * https://web.archive.org/web/20080308073422/http://www2.hursley.ibm.com/decimal/DPDecimal.html
+ * originally written in the (obselete) language Rexx.
```
Googulator, line 12: typo: obselete -> obsolete

PR #497, opened 2025-02-04
Improve support for pre-python 3.12 f-strings.

stikonas:


rootfs.py
```diff
@@ -55,7 +55,7 @@ def create_configuration_file(args):
         config.write(f"CONFIGURATOR={args.configurator}\n")
         if not args.external_sources:
             if args.mirrors:
-                config.write(f"MIRRORS=\"{" ".join(args.mirrors)}\"\n")
+                config.write(f'MIRRORS=\"{" ".join(args.mirrors)}\"\n')
```
fosslinux, line 58: we can remove the escapes for the double quotes then too

fosslinux (approved):

PR #498, opened 2025-02-05
Replace byacc with oyacc

fosslinux:
byacc actually has an awk script to generate .c files. Doesn't look trivial to rewrite in C.

Thanks to lanodan on #bootstrappable IRC, oyacc actually works fine instead, so let's replace byacc with oyacc. Builds bash correctly too.


parts.rst
```diff
@@ -322,15 +322,23 @@ we skip them.
 
 The ``cp`` in this stage replaces the ``mescc-tools-extra`` ``cp``.
 
-byacc 20240109
-==============
+oyacc 6.6
```
Googulator, line 325: heirloom-devtools's description later still references Berkeley Yacc.

fosslinux: fixed

stikonas (approved):

fosslinux:
Just want to check @Googulator is happy with + doesn't see any problems with this, as he added byacc originally.

stikonas:
> Just want to check @Googulator is happy with + doesn't see any problems with this, as he added byacc originally.

@Googulator did leave a comments earlier that you resolved. So perhaps we can assume that it's fine to merge.

Googulator (approved):
Yes, forgot about this, but it is fine.

PR #504, opened 2025-02-08
Move coreutils 6.10 to VCS + Gnulib

fosslinux:
Some files were a bit suspicious throughout lib/; this is better in terms of pregenerated files.

Namely, a lot of files in `lib/` had weird "DO NOT EDIT" headers. Some of them were slightly different to the gnulib import, and I can't find any script that generated them. This seems to be a bit more robust.

Not trivial, review would be good


stikonas (approved):
I guess fine.

The only difference between repo and tarball I find is that header "DO NOT EDIT" and GPL version updated from 2 to 3...

fosslinux:
eh, yeah, i didn't actually check the diff between the `lib/` directories in each. This is probably not needed then, just looks like someone has added a DO NOT EDIT header in the final tarball.

PR #506, opened 2025-02-08
Perl fixes

fosslinux:
more pregenerated files, this one has a bit more


steps/perl-5.6.2/pass1.sh
```diff
@@ -36,4 +44,10 @@ src_prepare() {
 
 src_compile() {
     make -j1 PREFIX="${PREFIX}"
+
+    cd lib/unicode
+    # We don't use the Makefile because $0 is encoded and it uses ./mktables.PL
+    # whereas the original uses mktables.PL
+    ../../miniperl -I../../lib mktables.PL
```
stikonas, line 51: Some of the files in `lib/unicode/Is` only have 444 permissions. Do we want to chmod them just in case?

stikonas (approved):

steps/perl-5.6.2/pass1.sh
fosslinux: Do you mean chmod-ing them after the regeneration?

I'm a bit confused how mktables.PL is even capable of regenerating those files. Perhaps our tar doesn't preserve permissions properly.

stikonas: I thought before. So that we have permissions to overwrite.

I guess it works because we run as root (or at least as fake namespaced root).

stikonas: Oh I see we remove the files before in `src_prepare`. Then I guess nothing to do here...

PR #508, opened 2025-02-10
Binutils regenerations

fosslinux:
This one was painful......


steps/binutils-2.30/pass1.sh
```diff
@@ -1,73 +1,123 @@
...
+    rm zlib/contrib/masmx86/*.obj \
+        zlib/contrib/infback9/inffix9.h \
+        zlib/contrib/blast/test.pk \
+        zlib/contrib/puff/zeros.raw \
+        zlib/contrib/dotzlib/DotZLib.chm \
+        zlib/crc32.h zlib/inffixed.h
+
+    # Regenerate crc table in libiberty/crc32.c
+    cd libiberty
+    sed -n -e '38,65p' crc32.c > crcgen.c
```
stikonas, line 71: Hmm, this might be annoying if we upgrade binutils to another version... Line numbers might change.

Maybe it would be safer to do pattern matching? sed can do something like

`sed '/start_pat/,/end_pat/d'`

stikonas (approved):
Hmm, CI failed though.

PR #512, opened 2025-02-16
xz: update and regen files

fosslinux:
5.4.1 -> 5.6.4

regenerate/delete files (this would have detected the infamous xz backdoor)


stikonas (approved):

steps/xz-5.6.4/sources
```diff
@@ -0,0 +1 @@
+https://github.com/tukaani-project/xz/releases/download/v5.6.4/xz-5.6.4.tar.bz2 176d510c30d80a23b8050bbc048f2ecaacb823ae48b6821727ed6591f0df9200
```
stikonas, line 1: @fosslinux by the way, do we know use https sources or still use http before openssl?

I tried running without mirrors and it obviously failed.

PR #517, opened 2025-03-22
Change mescc to use a unified libc C file

fosslinux:
Significantly speeds up libc compilation.


steps/mes-0.27/pass1.kaem
```diff
@@ -96,180 +96,188 @@ alias mescc="${MES} -e main ${BINDIR}/mescc.scm -D HAVE_CONFIG_H=1 -I include -I
...
-mescc lib/mes/oputs.c
 mescc lib/mes/globals.c
-mescc lib/stdlib/exit.c
-mescc lib/linux/${MES_ARCH}-mes-mescc/_exit.c
-mescc lib/linux/${MES_ARCH}-mes-mescc/_write.c
-mescc lib/stdlib/puts.c
-mescc lib/string/strlen.c
-catm ${LIBDIR}/${MES_ARCH}-mes/libc-mini.a __init_io.o eputs.o oputs.o globals.o exit.o _exit.o _write.o puts.o strlen.o
-catm ${LIBDIR}/${MES_ARCH}-mes/libc-mini.s __init_io.s eputs.s oputs.s globals.s exit.s _exit.s _write.s puts.s strlen.s
+catm libc-mini-1.c \
```
stikonas, line 100: can we not just have libc-mini.c and globals.c. Or is the order of object files somehow important? I don't think it should be, hex2 should be able to link anyway.

stikonas: Perhaps even  catting globals.c into both libc-mini.c and libc.c could make sense? One file would be double compiled but it seems that we'll save just as much time by removing multiple start overhead.

fosslinux: > Perhaps even catting globals.c into both libc-mini.c and libc.c could make sense? One file would be double compiled but it seems that we'll save just as much time by removing multiple start overhead.

Yeah.. let's just do that. I did it originally for `libmescc` but it probably makes more sense to do the same thing there

PR #525, opened 2025-06-03
Make download-distfiles.sh read from steps/manifest

alganet:
This brings the shell script closer to what get_packages() at lib/generator.py does.

It allows for downloading only what is specified in the manifest.

---

This is useful for partial builds or custom manifests. I also removed some bashisms (arrays, `[[`) to improve portability to shells other than bash.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,5 +6,6 @@
 target/
 kernel
 distfiles/
+mirrorstate/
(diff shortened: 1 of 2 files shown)


fosslinux (approved):
LGTM. Thanks!!

PR #529, opened 2025-07-18
Define "TARGET" variable to avoid hardcoding "i386-unknown-linux-musl" everywhere

eduardosm:


fosslinux (approved):
This makes a lot of sense. Thank you!

PR #531, opened 2025-08-18
Use i686-unknown-linux-musl target instead of i386-unknown-linux-musl

eduardosm:
The Linux kernel is already built with CONFIG_M686=y, which suggests that the bootstrap is not supported on anything older thant i686. So, use i686 on userspace too. Some software, such as gettext, cmake or elfutils, has trouble building for i386 due to missing atomic intrinsics.

For example, building gettext 0.26 would fail with:

```
/usr/lib/i386-unknown-linux-musl/gcc/i386-unknown-linux-musl/13.3.0/../../../../../i386-unknown-linux-musl/bin/ld: /build/gettext-0.26_1/gettext-0.26/build/gettext-tools/gnulib-lib/.libs/libgettextlib.so: undefined reference to `__atomic_compare_exchange_4'
```


Googulator:
This has caused a regression on bare metal due to musl still linking /lib/ld-linux.so.2 to the old i386 path, which no longer exists.

PR #535, opened 2025-09-03
Update final gcc to 15.2.0

eduardosm:


eduardosm:
libtool-2.2.4 package, which shouldn't be affected by this PR, with wrong checksum.

diffoscope: [libtool-diff.html](https://github.com/user-attachments/files/22196507/libtool-diff.html)

PR #537, opened 2025-09-07
Remove additional generated files in libtool-2.2.4

eduardosm:
Most of the times, `libtoolize.in` was not being regenerated by `bootstrap`. However, there was at least one spurious case where it got regenerated, leading to a package with a different checksum. By deleting it before running `bootstrap`, it will be always regenerated, ensuring a consistent result.

Hopefully, this fixes the spurious non-reproducibility found in https://github.com/fosslinux/live-bootstrap/pull/535#issuecomment-3263729677


stikonas:
Hmm, should we also delete `tests/package.m4` from libtool-2.4.7? defs.in doesn't seem to be present there.

stikonas:
Also needs rebasing now.

fosslinux:
Yes, would be good to cross-check these against libtool-2.4.7. Otherwise this seems sensible.

PR #541, opened 2025-09-19
Update mes to 0.27.1

stikonas:
Somehow riscv bootstrap stopped a bit earlier than I thought it would (just after tcc-mes was built). We probably need to update tcc-0.9.26 snapshot a bit but it's experimental platform anyway.

This also brings in new M2-Planet 1.13.1 with various new features. mes 0.27.1 is not fully compatible with it, so there is a small patch to mes that is already upstreamed.


fosslinux (approved):
Yeah, looks good, thanks

PR #545, opened 2025-10-03
Temporarily change gnulib git repo to github

fosslinux:
We have had a stupidly large number of problems with savannah's git service recently, especially when cloning gnulib. Let's move gnulib to being from the GitHub mirror, at least for the time being.


fosslinux:
actually found a better idea

PR #546, opened 2025-10-03
Better git repo handling

fosslinux:
This really should have been done a while ago. Instead of detecting the type of distfile based on the protocol, explicitly specify it. This allows us to do things like use https:// for git repositories without weird hacks.

(This also fixes the problems with savannah, because it doesn't use the git:// protocol anymore, but https://, which works fine)


fosslinux:
Oh of course. will fix later

PR #579, opened 2026-03-20
fix(payload): port raw external.img container

vxtls:
just leave it here, not tested yet

### DIFF
--- a/Payload_img_design.md
+++ b/Payload_img_design.md
@@ -1,0 +1,106 @@
+# live-bootstrap
+
+This repository uses [`README.rst`](./README.rst) as the canonical main documentation.
+
+## Kernel-bootstrap raw `external.img`
+
+`external.img` is a raw container disk used in kernel-bootstrap mode when
+`--external-sources` is set and `--repo` is unset.
+
+### Why not put everything in the initial image?
+
+In kernel-bootstrap mode, the first boot image is consumed by very early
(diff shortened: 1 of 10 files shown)


vxtls:
Testing right now, once success i will open this PR

Payload_img_design.md
fosslinux: Let's move the contents of this into `DEVEL.md` or `README.rst` as appropriate.

Payload_img_design.md
```diff
@@ -0,0 +1,106 @@
...
+
+In kernel-bootstrap mode, the first boot image is consumed by very early
+runtime code before the system reaches the normal bash-based build stage.
+That early stage has tight assumptions about memory layout and file table usage.
+
+When too many distfiles are packed into the initial image, those assumptions can
+be exceeded, which leads to unstable handoff behavior (for example, failures
+around the Fiwix transition in QEMU or on bare metal).
+
+So the design is intentionally split:
```
fosslinux: remove "intentionally", it is obviously intentional

fosslinux:
ok, I need to think a bit more about this, I didn't realise that the distfiles get copied onto the main disk now. This won't do as is then, because there are more things that could be in `external.img`, like `/external/repo`... hold off on this until I think a bit more about design, unless you have a better alternative design already in mind?

PR #583, opened 2026-07-27
riscv64: pin tcc 0.9.26-1157-gdd46e018 (1147's tcc-mes cannot build through)

### LINKED ISSUE #582: riscv64: pinned tcc-mes (0.9.26-1147) cannot build through; the checksum set mixes eras
The riscv64 checksum set describes a chain that cannot be built through
as pinned.

What does reproduce: building tcc-mes from the pinned
tcc-0.9.26-1147-gee75a10c with the mes-0.27.1 MesCC toolchain yields
the pinned binary byte-exactly (sha256 3e2d226b...). But that binary
cannot take the next step, for two independent reasons in revision
1147:

1. riscv64-gen.c's immediate-range assert,
   assert(!((imm + (1 << 11)) >> 12)), false-positives on every
   negative 12-bit immediate when tcc was built by MesCC: MesCC-riscv64
   evaluates the uint32 addition as a full 64-bit add without
   re-truncating mod 2^32 (long-standing MesCC behaviour, identical in
   mes 0.26.1 and 0.27.1; reported to bug-mes with a minimal
   reproducer). The addi sp,sp,-16 of the first function epilogue
   trips it, so compiling even void f(){} dies. Revision 1157 casts
   the expression to (uint32_t) (riscv64-gen.c:123,130), which MesCC
   evaluates correctly.

2. 1147's assembler has no .option directive (added in the 1157 era),
   and mes-0.27.1's riscv64 crt1.c startup asm uses
   .option push/norelax/pop -- crt1.o cannot assemble at all.

Meanwhile the downstream riscv64 pins are outputs of the LATER
revision: a self-rebuilt tcc-0.9.26-1157-gdd46e018 compiles
mes-0.27.1's crt1.c to the pinned crt1.o cc417e9d... byte-exactly.
Commit 865b9aea ("Update mes to 0.27.1") refreshed only the tcc-mes
hash, so the set mixes eras, and no buildable path leads from the
pinned tcc-mes to the artifacts the later entries record.

One nuance for regeneration: the 1157 tcc built BY MesCC (generation
1) takes benign large-offset fallbacks at its now-non-fatal range
checks and produces a different but working crt1.o; its self-rebuild
has correct semantics and reproduces cc417e9d.... Byte verification
should target the self-rebuilt compiler's outputs -- which is what the
recorded downstream pins already are.

A CI job that re-checks what master pins today and evaluates the 1147
and 1157 range-check guards verbatim under MesCC (host gcc as passing
control) is live-bootstrap-riscv64-stale-pins.yml at
https://github.com/JasonGross/bootstrap-chain-bug-reproducers

The companion pull request bumps the riscv64 pin to 1157.

---
Authorship note: this issue was researched, written, and posted by
Claude (Anthropic's Fable 5 model), working on Jason Gross's behalf.

JasonGross:
Fixes #582.

riscv64-only bump of the tcc-0.9.26 step to tcc-0.9.26-1157-gdd46e018.
x86 keeps 1147 and its checksums; the step's fileopen simple-patches
apply unchanged to 1157's tcctools.c. Why 1147 cannot build through on
riscv64 -- a MesCC-tripped range assert that 1157 casts to (uint32_t),
and a missing .option directive that mes-0.27.1's crt1.c needs -- is
in the issue.

Contents:

- steps/tcc-0.9.26/sources: add the 1157 tarball from the same
  lilypond.org directory (sha256 3748c0aa...).
- steps/tcc-0.9.26/pass1.kaem: select the 1157 tree when ARCH is
  riscv64, in the file's existing arch-conditional style.
- tcc-0.9.26.riscv64.checksums: /usr/bin/tcc-mes becomes bff3b8f3...
  (1020608 bytes), the MesCC-built 1157 binary. Measured by a replay
  of this step's recipe; the same replay reproduces the current pin
  3e2d226b... byte-exactly when pointed at 1147, which is our evidence
  the recipe is faithful. We could not run the full in-tree bootstrap
  here.

The downstream entries are left as recorded: they are already the 1157
lineage's outputs. The one we verified byte-exactly is crt1.o
cc417e9d... -- our replay's self-rebuilt 1157, the recorded pin, and
the tcc that Guix's riscv64 bootstrap builds all agree on it. A run
with UPDATE_CHECKSUMS=True would confirm the whole file from one
lineage; we changed only what we measured rather than guess the rest.

Authorship note: this pull request was researched, written, and posted
by Claude (Anthropic's Fable 5 model), working on Jason Gross's behalf.


fosslinux (changes requested):
this doesn't make sense to me, why have different versions for riscv64 and everything else?

JasonGross:
Yeah, that's a good point, I guess the LLM was trying to code "defensively" by only changing things it tested.  I will fix.

stikonas:
New mes should be able to build tcc 0.9.28pre (mob) directly as far as I've heard.

stikonas:
or if we can't achieve direct to 0.9.28, it is still better to sort this in tcc fork (e.g. with some ifdef sections) rather than live-bootstrap

PR #586, opened 2026-08-24
Fix typos across the repository.

siraben:


stikonas (approved):

stikonas:
CI is broken though... We need to fix that first. Otherwise I can't really tell whether any checksums change, e.g. due to FGREP type fix.

